PR #9, opened 2023-03-16
Update quickstart instructions to account for CodeFlare operator

anishasthana:
cc @MichaelClifford

### DIFF
--- a/Quick-Start.md
+++ b/Quick-Start.md
@@ -19,10 +19,6 @@ This Quick Start guide assumes that you have administrator access to an OpenShif
 
 If you want to run GPU enabled workloads, you will need to install the [Node Feature Discovery Operator](https://github.com/openshift/cluster-nfd-operator) and the [NVIDIA GPU Operator](https://github.com/NVIDIA/gpu-operator) from the OperatorHub. 
 
-### Helm
-
-We also assume that you have Helm 3 installed in your working environment. You can find instructions to install it [here](https://helm.sh/docs/intro/install/).   
-
 
 ## Clone the demo code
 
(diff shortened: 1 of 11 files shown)


anishasthana:
@KPostOffice @accorvin @MichaelClifford ptal

I'm thinking we can have a follow-up PR later to move kuberay into this repo.

MichaelClifford (approved):
:rocket: 
LGTM! Thanks @anishasthana

codeflare-stack-kfdef.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: kfdef.apps.kubeflow.org/v1
+kind: KfDef
```
accorvin, line 2: Should this kfdef include the minimal odh stuff as well (dashboard and notebook controller?)

anishasthana: Good question... The way I see it, since we're not part of odh-core we should just create a kfdef managing "our" components. I could update the quickstart guide to say something along the lines of:
1) Create odh-core kfdef
2) Create Codeflare-stack kfdef

That being said, if there's a strong preference for us to include the odh-core kfdef components in here, I'm happy to make the update.

accorvin: Our instructions should assume you know nothing about ODH. Given that, one of our users isn't going to know how to get the odh core components in the current state.

 If we go with the plan for telling a user to first apply an odh core kfdef in addition to our kfdef, then we should link them to a suitable odh core kfdef. If one doesn't exist then we should include the odh core components in our kfdef.

anishasthana: I've updated the instructions, wdyt?

Quick-Start.md
```diff
@@ -33,80 +29,30 @@ git clone https://github.com/project-codeflare/codeflare-sdk
 cd codeflare-sdk
 ```
 
-## Install the CodeFlare Stack
+## Install the CodeFlare Operator
 
-Fist we will use [Helm](https://helm.sh/) to install MCAD. (Make sure you have Helm 3 or above).
+First, install the CodeFlare operator from the operatorhub. The default settings for the operator will suffice.
```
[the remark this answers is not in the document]
anishasthana: ODH is currently recommending we stay away from subscriptions. They are in the process of figuring out a better dependency strategy -- once that's solidified we can update this quickstart/included artifacts to conform to their standards.

codeflare-stack-kfdef.yaml
accorvin: Yes, much better!

PR #52, opened 2023-06-13
Updating quickstart guide

Fiona-Waters:
## Description
Updating Quick-Start guide to include creation of required instascale-ocm-secret.

## Merge criteria:

### DIFF
--- a/Quick-Start.md
+++ b/Quick-Start.md
@@ -53,6 +53,8 @@ If you want to run GPU enabled workloads, you will need to install the [Node Fea
     oc create ns opendatahub
     ```
 
+1. Create a secret in the opendatahub namespace containing your ocm token. You can find your token [here](https://console.redhat.com/openshift/token). Navigate to Workloads -> secrets in the Openshift Console. Click Create and choose a key/value secret. Secret name: instascale-ocm-secret Key: token, Value: <ocm token> and click create.
+
 1. Apply the odh-core kfdef with this command:
 
     ```bash


dimakis (approved):
lgtm

accorvin:
@anishasthana I'd like your input on this. To me, this file should not always be required. It should only be necessary on OCM-provisioned clusters, and I'd expect the file to automatically be placed so that the user doesn't have to worry about things. If that's not the case, there might be a bug.

anishasthana (changes requested):
Agreed on it not always being needed. That being said, there's no way for us to get the OCM secret from within the cluster. Users _have_ to create this secret manually after getting their token from OCM.

We should add a separate set of instructions for OSD

Fiona-Waters:
> Agreed on it not always being needed. That being said, there's no way for us to get the OCM secret from within the cluster. Users _have_ to create this secret manually after getting their token from OCM.
> 
> We should add a separate set of instructions for OSD

Apologies but I am not sure what these separate instructions should be? Happy to make whatever changes necessary.

accorvin:
> That being said, there's no way for us to get the OCM secret from within the cluster. Users have to create this secret manually after getting their token from OCM.

TIL. We should talk to the OCM team about what can be done here.

Fiona-Waters:
@anishasthana @accorvin In terms of this current issue, do we want to add this to the quick start guide, or should we wait until we are clearer on what will be needed? Thanks

Quick-Start.md
```diff
@@ -78,6 +78,10 @@ At this point you should be able to go to your notebook spawner page and select
 
 You can access the spawner page through the Open Data Hub dashboard. The default route should be `https://odh-dashboard-<your ODH namespace>.apps.<your cluster's uri>`. Once you are on your dashboard, you can select "Launch application" on the Jupyter application. This will take you to your notebook spawner page.
 
+
+### **Using an Openshift Dedicated or ROSA Cluster**
+If you are using an Openshift Dedicated or ROSA Cluster you will need to create a secret in the opendatahub namespace containing your ocm token. You can find your token [here](https://console.redhat.com/openshift/token). Navigate to Workloads -> secrets in the Openshift Console. Click Create and choose a key/value secret. Secret name: instascale-ocm-secret, Key: token, Value: < ocm token > and click create.
```
anishasthana, line 83: Can you paste a screenshot or something similar here? it would make the steps more obvious to someone who is not familiar with the console/ocm/secret creation

anishasthana (approved):
/lgtm

PR #56, opened 2023-06-14
add scc that forces ray pods to run as user 1000

KPostOffice:
## Description
fixes: https://github.com/project-codeflare/codeflare-sdk/issues/129

## How Has This Been Tested?
I tested this on my local cluster by explicitly creating the SCC, I will retest after completely removing kuberay

## Merge criteria:

### DIFF
--- a/ray/operator/scc/ray_operator_scc.yaml
+++ b/ray/operator/scc/ray_operator_scc.yaml
@@ -1,0 +1,11 @@
+kind: SecurityContextConstraints
+apiVersion: security.openshift.io/v1
+metadata:
+  name: run-as-ray-user
+seLinuxContext:
+  type: MustRunAs
+runAsUser:
+  type: MustRunAs
+  uid: 1000
+users:
+  - 'system:serviceaccount:$(namespace):kuberay-operator'
--- a/ray/operator/base/kustomization.yaml
+++ b/ray/operator/base/kustomization.yaml
@@ -22,6 +22,7 @@ resources:
 - ../rbac
 - ../manager
 - ../prometheus
+- ../scc
 
 images:
 - name: kuberay/operator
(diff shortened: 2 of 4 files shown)


ray/operator/scc/ray_operator_scc.yaml
```diff
@@ -0,0 +1,10 @@
+kind: SecurityContextConstraints
+metadata:
+  name: run-as-ray-user
+seLinuxContext:
+  type: MustRunAs
+runAsUser:
+  type: MustRunAs
+  uid: 1000
+users:
+  - 'system:serviceaccount:opendatahub:kuberay-operator'
```
KPostOffice, line 10:
```suggestion
  - 'system:serviceaccount:<deployed-namespace>:kuberay-operator'
```

I need help getting the namespace to be based on where the kuberay operator is actually deployed.

KPostOffice: Just using the var ref works fine. It doesn't have to replace the whole value as I thought that it did.

MichaelClifford (approved):

MichaelClifford:
I ran this on our teams cluster with ODH and Codeflare and it successfully overcomes the issue pointed out in https://github.com/project-codeflare/codeflare-sdk/issues/129

anishasthana:
Heh, I wanted to confirm the scc isn't a "bad" idea. @astefanutti can you also take a look at this tomorrow?

astefanutti:
I can't think of any cons for using SCC.

It's not clear to me how that issue has something to do with the kuberay operator SA, rather than the SA that's used to run the Ray pods. From the referenced issues, it seems more related to the later than the former, which is the one added to the SCC.  That's probably my lack of the Kuberay internals, but I'd be curious to find out what I'm actually missing here.

Anyhow, it fixes the issue, and ideally, it'll be tackled in Ray eventually, so the images can run with any users.

PR #89, opened 2023-08-22
Defer retrieving GetRayJobId in test

sutaakar:
GetRayJobId call needs to be deferred to make sure that job id is stored by KubeRay cluster.

## Description

TestRayCluster test was failing on log storage line. Log storage function was deferred, however its arguments are resolved in realtime, causing GetRayJobId to be resolved before KubeRay operator had a chance to store job id in RayJob CR. 

Moving the whole log storage line into separate function makes sure that RayJob CR is fully initialized.

## How Has This Been Tested?

Manually.

## Merge criteria:

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -3,7 +3,7 @@ module github.com/opendatahub-io/distributed-workloads/tests/new-tests
 require (
 	github.com/onsi/gomega v1.27.8
 	github.com/openshift/api v0.0.0-20230718161610-2a3e8b481cec
-	github.com/project-codeflare/codeflare-operator v0.1.1-0.20230817124305-be0dd48b43b4
+	github.com/project-codeflare/codeflare-operator v0.1.1-0.20230823091637-6f7d6b1194dd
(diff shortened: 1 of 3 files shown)


winklerm (approved):

sutaakar:
> That makes me think whether we should update the Ray API helper function signatures to take the Job namespace and name, instead of the Job id, and do the retrieval from within the helper function.

@astefanutti The problem is that there are two options of submitting the job - CR or REST API. In case of REST API there is no CR created, so helper wouldn't be able to obtain job id in that case from namespace and name.

astefanutti:
> > That makes me think whether we should update the Ray API helper function signatures to take the Job namespace and name, instead of the Job id, and do the retrieval from within the helper function.
> 
> @astefanutti The problem is that there are two options of submitting the job - CR or REST API. In case of REST API there is no CR created, so helper wouldn't be able to obtain job id in that case from namespace and name.

Ah right. So we could (re-)add the `WriteRayJobLogs` function in `ray.go` that would handle the RayJob case, and delegate to `WriteRayJobAPILogs` in `ray_api.go`. WDYT?

PR #125, opened 2023-09-21
Upgrade to new operator design

astefanutti:
## Description

Upgrade the CodeFlare operator to v1.0.0-rc.1 that contains the new operator design.

Closes #112.
Closes #113.

PR #127, opened 2023-09-22
Bind Ray ClusterRole to the new operator ServiceAccount

astefanutti:
Fix a missing update from #125, to bind the Ray ClusterRole to the CodeFlare operator ServiceAccount.

It'll be simplified once project-codeflare/multi-cluster-app-dispatcher#635 is implemented.


sutaakar:
/lgtm
Though it doesn't have any effect yet as operator still use cluster edit role.

astefanutti:
> /lgtm
> Though it doesn't have any effect yet as operator still use cluster edit role.

It doesn't seem like Kuberay resources are aggregated to the `edit` ClusterRole, `kubectl get clusterrole edit -o yaml | grep ray` doesn't return anything. If that is correct, then it should fixed the RBAC issue reported by @jbusche.

dimakis (approved):
LGTM

PR #142, opened 2023-10-16
Add Routes to Ray cluster role

astefanutti:
This PR follows up project-codeflare/codeflare-operator#316, and adds permissions on Routes, that's currently required by the SDK to expose the dashboard endpoints of provisioned Ray clusters.


anishasthana (approved):
/lgtm

PR #148, opened 2023-10-20
Remove job that adds needs-triage label

### LINKED ISSUE #147: Update auto-add-issues GitHub Action
## Description
The [current workflow](https://github.com/opendatahub-io/distributed-workloads/blob/05d144b3456363e35b6a965cda736c72dd9cad19/.github/workflows/auto-add-issues.yaml) attempts to add a `needs-triage` label to new issues created.

This label has been removed as we are now using `priorities` in the CodeFlare Sprint Board instead.

## Changes required
Update the workflow

ChristianZaccaria:
Closes #147 

## Description

Removed the job and updated workflow as we now make use of `priorities` instead.

## Open Question
- Should we add a different label such as a priority?
- Or, add no labels. To be added during grooming.


sutaakar:
Not relevant any more, closing

PR #152, opened 2023-11-07
Use busybox from quay.io

jiripetrlik:
Use busybox from quay.io instead of from Docker hub. Version should stay the same.

### DIFF
--- a/tests/integration/ray_test.go
+++ b/tests/integration/ray_test.go
@@ -319,7 +319,7 @@ func createRayCluster(test support.Test, namespace string, mnist *corev1.ConfigM
 							InitContainers: []corev1.Container{
 								{
 									Name:    "init-myservice",
-									Image:   "busybox:1.28",
+									Image:   "quay.io/project-codeflare/busybox:1.36",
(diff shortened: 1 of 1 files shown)


tests/integration/ray_test.go
```diff
@@ -319,7 +319,7 @@ func createRayCluster(test support.Test, namespace string, mnist *corev1.ConfigM
 							InitContainers: []corev1.Container{
 								{
 									Name:    "init-myservice",
-									Image:   "busybox:1.28",
+									Image:   "quay.io/project-codeflare/busybox:1.28",
```
dimakis, line 322: Is there a reason we're using `1.28` and not `1.36` which appears to be the latest one we've got in quay?

dimakis: (quay page)[https://quay.io/repository/project-codeflare/busybox?tab=tags]

jiripetrlik: @dimakis Yes, true you are right. Updated to 1.36.

jiripetrlik:
Replaced by https://github.com/opendatahub-io/distributed-workloads/pull/153

PR #153, opened 2023-11-08
Remove busybox container from tests

jiripetrlik:
It seems that busybox container is no longer needed.


dimakis (commented):
have we tested this?

jiripetrlik:
I've run TestRayJobSubmissionRest locally. Also PR check seems to fail only on MCAD jobs and MCAD jobs are expected to fail am I right @sutaakar ?

sutaakar:
Right, merging

PR #154, opened 2023-11-13
Temporary fix the test because of RHODS-12857

jiripetrlik:
Change assert to temporary fix the test for 2.4.


sutaakar (approved):
Approved as a workaround, will be replaced in near future.

PR #155, opened 2023-11-13
Do not save logs when RayJob status is missing

jiripetrlik:
It is because of RHODS-12857. Utilities which save the logs require RayJob status which may not be available.

### DIFF
--- a/tests/integration/ray_test.go
+++ b/tests/integration/ray_test.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"testing"
 
+	"github.com/onsi/gomega"
 	. "github.com/onsi/gomega"
 	support "github.com/project-codeflare/codeflare-operator/test/support"
 	rayv1alpha1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1alpha1"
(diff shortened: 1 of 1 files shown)


tests/integration/ray_test.go
```diff
@@ -108,13 +108,16 @@ func TestRayCluster(t *testing.T) {
 	test.Eventually(support.Route(test, dashboardRoute.Namespace, dashboardRoute.Name), support.TestTimeoutLong).
 		Should(WithTransform(support.ConditionStatus(routev1.RouteAdmitted), Equal(corev1.ConditionTrue)))
 
-	// Retrieve dashboard hostname
-	dashboard, err := test.Client().Route().RouteV1().Routes(namespace.Name).Get(test.Ctx(), dashboardRoute.Name, metav1.GetOptions{})
-	test.Expect(err).NotTo(HaveOccurred())
-	dashboardHostname := dashboard.Status.Ingress[0].Host
```
astefanutti, line 114: Maybe an if statement the checks the status would be better than commenting out the code?

jiripetrlik: Thank you @astefanutti for good tip. I've created a function which checks if RayJob status is present first. If yes it saves the logs if not it will skip saving the log.

PR #173, opened 2024-06-18
Add Fine-Tune LLMs with Ray and DeepSpeed example

astefanutti:


examples/ray-finetune-llm-deepspeed/ray_finetune_llm_deepspeed.ipynb
```diff
@@ -0,0 +1,209 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4d340f86-1a04-48d3-a5e7-067faecfc17a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "cluster = Cluster(ClusterConfiguration(\n",
```
Bobbins228, line 56: Should we include the same comments that we do for our regular demo notebooks for Token Authentication and Cluster Configuration?

examples/ray-finetune-llm-deepspeed/ray_finetune_llm_deepspeed.ipynb
```diff
@@ -0,0 +1,209 @@
...
+    "    num_workers=7,\n",
+    "    min_cpus=16,\n",
+    "    max_cpus=16,\n",
+    "    head_cpus=16,\n",
+    "    min_memory=128,\n",
+    "    max_memory=256,\n",
+    "    head_memory=128,\n",
+    "    head_gpus=1,\n",
+    "    num_gpus=1,\n",
+    "    image=\"quay.io/rhoai/ray:2.23.0-py39-cu121-torch\",\n",
```
sutaakar, line 68: Is there any reason to use `quay.io/rhoai/` instead of `quay.io/project-codeflare/ray`?

examples/ray-finetune-llm-deepspeed/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+> [!IMPORTANT]
+> This example has been tested with the configurations listed in the [validation](#validation) section.
+> Its configuration space is highly dimensional, with application configuration tighly coupled to runtime / hardware configuration.
+> It is your responsibility to adapt it, and validate it works as expected, with your configuration(s), on your target environment(s).
+
+## Requirements
+
+* An OpenShift cluster with OpenShift AI (RHOAI) 2.10+ installed:
+  * The `codeflare`, `dashboard`, `ray` and `workbenches` components enabled;
```
sutaakar, line 14: I guess Kueue component should be enabled too, right?

astefanutti: It's not really mandatory for that example, so I'd be inclined to keep it the simplest.

examples/ray-finetune-llm-deepspeed/ray_finetune_llm_deepspeed.ipynb
astefanutti: The intent is too have examples for Distributed Workloads / OpenShift AI, which is larger than Codeflare. Also we'll probably move away from `quay.io/project-codeflare/ray` for the product.

examples/ray-finetune-llm-deepspeed/ray_finetune_llm_deepspeed.ipynb
```diff
@@ -0,0 +1,209 @@
...
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "# Create the training and evaluation datasets.\n",
+    "# This can be run only once.\n",
+    "!{sys.executable} -m pip install datasets\n",
+    "import create_dataset\n",
+    "create_dataset.main()"
```
sutaakar, line 31: Thinking whether it would have sense to have the datasets already available in repository, instead of creating them. That would allow us to remove this step and get rid of another external dependency.

If the dataset is https://huggingface.co/datasets/openai/gsm8k then license seems permissive to do so (MIT). Also dataset size is small enough to be kept in Git.
WDYT?

sutaakar: Of course the python script to create and adjust dataset should be still kept somewhere.

examples/ray-finetune-llm-deepspeed/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM rayproject/ray:2.23.0-py39-cu121
```
sutaakar, line 1: Would it have sense to keep this Dockerfile among others in https://github.com/opendatahub-io/distributed-workloads/tree/ray-finetune-llm-deepspeed/images ?

examples/ray-finetune-llm-deepspeed/ray_finetune_llm_deepspeed.ipynb
astefanutti: I think that's a good idea. And set the `HF_DATASETS_CACHE` environment variable.

examples/ray-finetune-llm-deepspeed/ray_finetune_llm_deepspeed.ipynb
sutaakar: ok, thanks

examples/ray-finetune-llm-deepspeed/README.md
sutaakar: ok, sounds good
if Kueue will become mandatory for RHOAI DW then adjustments would be needed

examples/ray-finetune-llm-deepspeed/Dockerfile
astefanutti: Yes, I've moved it under the `images` directory.

examples/ray-finetune-llm-deepspeed/ray_finetune_llm_deepspeed.ipynb
astefanutti: Updated, with the GSM8k dataset checked in the `datasets` directory.

The `HF_DATASETS_OFFLINE` environment variable can be set for truly offline mode.

examples/ray-finetune-llm-deepspeed/ray_finetune_llm_deepspeed.ipynb
astefanutti: I've added some comments, slightly more concise than from the one upstream.

sutaakar (approved):

PR #175, opened 2024-06-19
Move TestMnistRay from integration tests repo to ODH DW repo

sutaakar:
## Description

As part of test reconciliation I am moving TestMnistRay from https://github.com/project-codeflare/integration-tests here. That way all the integration tests for DW stack are in one place.

## Merge criteria:

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,23 @@
 
 * Installed Go 1.21
 
-### Environment variables
+### Common environment variables
 
 * `CODEFLARE_TEST_OUTPUT_DIR` - Output directory for test logs
 * `CODEFLARE_TEST_TIMEOUT_SHORT` - Timeout duration for short tasks
 * `CODEFLARE_TEST_TIMEOUT_MEDIUM` - Timeout duration for medium tasks
 * `CODEFLARE_TEST_TIMEOUT_LONG` - Timeout duration for long tasks
+
+### Environment variables for Training operator test suite
+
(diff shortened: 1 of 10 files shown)


jiripetrlik (approved):

astefanutti (commented):
I would have suggested to add it to the `tests/ray` directory instead of `tests/odh` :)

abhijeet-dhumal:
> I would have suggested to add it to the `tests/ray` directory instead of `tests/odh` :)

@astefanutti Please correct me if I'm wrong, here I suppose currently this integration test is common for all Distributed workload components, so `test/odh` seems better option instead of `test/ray `:roll_eyes:

sutaakar:
I kept `odh` folder name there as the test is using Notebook for running SDK.
There will be (probably) more similar tests in future using Notebook from ODH/RHOAI and IMHO it would be good to keep them on single place.

astefanutti:
I see, the criteria to name that folder is the "target" environment.

My suggestion was to organise the tests by the main "component under test", like I assume it was the case for `tests/kfto`.

Following that logic, even if we use a notebook in the context of ODH, the main component under test is Ray, so you could have inside `tests/ray` some tests that use notebook / odh, and some other that directly use the SDK, like for local interactive use case, or using the RayCluster API directly.

Also, at some point, there will probably be some tests for the KFTO that use notebook as well.

PR #180, opened 2024-07-09
Fix GPU quota

jiripetrlik:


tests/odh/mnist_ray_test.go
```diff
@@ -66,7 +67,7 @@ func mnistRay(t *testing.T, gpus_enabled bool) {
 							},
 							{
 								Name:         corev1.ResourceName("nvidia.com/gpu"),
-								NominalQuota: resource.MustParse("0"),
+								NominalQuota: resource.MustParse(strconv.Itoa(numGpus)),
```
sutaakar, line 70: you can use `resource.MustParse(fmt.Sprint(numGpus))`

jiripetrlik: Done.

sutaakar (approved):

PR #186, opened 2024-07-15
Updated odh-generic image in notebook CR to avoid internal image registry…

abhijeet-dhumal:
## Description

- Updated image reference in notebook CR to avoid internal image registry usage 
- Update mnist script to download datasets

## Merge criteria:


README.md
```diff
@@ -28,9 +28,20 @@
 ### Environment variables for ODH integration test suite
 
 * `ODH_NAMESPACE` - Namespace where ODH components are installed to
-* `NOTEBOOK_IMAGE_STREAM_NAME` - Name of the image stream used for running Workbench
 * `NOTEBOOK_USER_NAME` - Username of user used for running Workbench
 * `NOTEBOOK_USER_TOKEN` - Login token of user used for running Workbench
+* `NOTEBOOK_IMAGE` - Image used for running Workbench
+* `CLUSTER_QUEUE_NAME` - Cluster queue resource name to be used while running tests
```
sutaakar, line 34: Where is `CLUSTER_QUEUE_NAME` specified or used?

tests/odh/resources/mnist.py
```diff
@@ -113,36 +127,99 @@ def configure_optimizers(self):
...
+                        # Unzip files
+                        with gzip.open(filePath, "rb") as f_in:
+                            with open(filePath.split(".")[:-1][0], "wb") as f_out:
+                                shutil.copyfileobj(f_in, f_out)
+                        # delete zip file
+                        os.remove(filePath)
+                    else:
+                        print(f"Failed to download file {file}")
+                except Exception as e:
+                    print(e)
```
sutaakar, line 169: This looks like a content which can be located in helper method (in codeflare-common ?) and reused.

tests/odh/resources/mnist.py
```diff
@@ -169,14 +246,14 @@ def test_dataloader(self):
...
-model = LitMNIST()
+model = LitMNIST(data_dir=local_mnist_path)
 
 print("GROUP: ", int(os.environ.get("GROUP_WORLD_SIZE", 1)))
 print("LOCAL: ", int(os.environ.get("LOCAL_WORLD_SIZE", 1)))
 
 # Initialize a trainer
 trainer = Trainer(
-    accelerator="has to be specified",
+    accelerator="auto",
```
sutaakar, line 256: Why did you do this change?

abhijeet-dhumal: Ohh, I was doing some testing and missed to change it back!

README.md
abhijeet-dhumal: Apologies, This is a mistake.. will remove it.
Thank you @sutaakar for reviewing !!

tests/odh/notebook.go
```diff
@@ -44,6 +42,7 @@ type NotebookProps struct {
 	ImageStreamName           string
 	ImageStreamTag            string
```
sutaakar, line 43: AFAIK these properties should be removed too, right?

tests/odh/resources/mnist.py
```diff
@@ -113,36 +124,98 @@ def configure_optimizers(self):
...
-                else:
-                    print(f"Failed to download file {file}")
-            except Exception as e:
-                print(e)
-        print(f"Downloaded MNIST dataset to... {downloadLocation}")
+        # download
+        print("Downloading MNIST dataset...")
+
+        if "{{.StorageBucketDefaultEndpointExists}}" != "true":
+            print("Using MNIST_DATASET_URL to download datasets...")
```
sutaakar, line 131: Thinking....
Why do we manually download dataset in this case?

abhijeet-dhumal: @sutaakar In case where `MNIST_DATSET_URL` provided is other than default one, similar to what we saw when we used `http://yann.lecun.com/exdb/mnist/` as default `MNIST_DATASET_URL` but it started showing 403 error each time while downloading datasets..

While considering downloading options for downloading MNIST datasets, I listed 3 usecases : 
1. Using MNIST_DATASET_URL
2. Using users Storage_bucket
3. using default pytorch functionality incase pre-requisites for above options doesn't fulfill

tests/odh/notebook.go
abhijeet-dhumal: Thank you karel for pointing this out !
I have removed it now as I missed these before ..

tests/odh/resources/mnist.py
sutaakar: 1. so basically it is a copy of dataset download from https://pytorch.org/vision/stable/_modules/torchvision/datasets/mnist.html#MNIST , right? WIth the exception that we don't use `http://yann.lecun.com/exdb/mnist/` but only the second link.

sutaakar: Makes me think why not just use `download=True` for MNIST and let the underlying layer take care of it.
AFAIK `MNIST_DATASET_URL` property was used initially as a mean to download datasets in disconnected environment. With usage of S3 bucket it loses a meaning.

abhijeet-dhumal: Thank you @sutaakar for bring this case to light, 
I was in assumption to reuse this parameter as this was already in use but now I have removed `MNIST_DATASET_URL` usecase as it is not needed anymore!

sutaakar (commented):
/lgtm

Srihari1192 (approved):

PR #242, opened 2024-10-10
Upstream Konflux pipelines

astefanutti:
Copy the latest versions of Konflux pipelines from downstream fork, so changes can be applied here.


sutaakar (commented):
/lgtm

Fiona-Waters (approved):
/approve

PR #244, opened 2024-10-15
Upload models trained using Training operator into s3 bucket

sutaakar:
## Description

Add optional possibility to upload models trained using Training operator into s3 bucket.
Such models can be then reused by other components like serving.

## How Has This Been Tested?

Manually by running kfto tests against OCP cluster, uploading trained models into AWS s3 bucket.

## Merge criteria:

### DIFF
--- a/images/util/mc-cli/Dockerfile
+++ b/images/util/mc-cli/Dockerfile
@@ -1,0 +1,10 @@
+FROM registry.access.redhat.com/ubi9:latest
+WORKDIR /app
+USER 1001
+
+RUN curl https://dl.min.io/client/mc/release/linux-amd64/mc -o /app/mc
+RUN chmod +x /app/mc
+ENV PATH="$PATH:/app"
+ENV MC_CONFIG_DIR="/app"
+RUN chown -R $USER:0 /app && \
+    chmod -R g+rwX /app
--- a/tests/kfto/core/environment.go
+++ b/tests/kfto/core/environment.go
@@ -30,14 +30,18 @@ const (
 	bloomModelImageEnvVar = "BLOOM_MODEL_IMAGE"
 	// The environment variable referring to image containing Stanford Alpaca dataset
 	alpacaDatasetImageEnvVar = "ALPACA_DATASET_IMAGE"
+	// The environment variable referring to image containing minio CLI
+	minioCliImageEnvVar = "MINIO_CLI_IMAGE"
 	// The environment variable for HuggingFace token to download models which require authentication
 	huggingfaceTokenEnvVar = "HF_TOKEN"
-	// The environment variable specifying existing namespace to be used for multiGPU tests
-	multiGpuNamespaceEnvVar = "MULTIGPU_NAMESPACE"
+	// The environment variable specifying existing namespace name to be used for tests
+	testNamespaceEnvVar = "TEST_NAMESPACE_NAME"
 	// The environment variable specifying name of PersistenceVolumeClaim containing GPTQ models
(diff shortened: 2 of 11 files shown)


tests/kfto/core/environment.go
```diff
@@ -30,14 +32,20 @@ const (
...
+	// The environment variable specifying existing namespace name to be used for tests
+	testNamespaceEnvVar = "TEST_NAMESPACE_NAME"
 	// The environment variable specifying name of PersistenceVolumeClaim containing GPTQ models
 	gptqModelPvcNameEnvVar = "GPTQ_MODEL_PVC_NAME"
 	// The environment variable referring to image simulating sleep condition in container
 	sleepImageEnvVar = "SLEEP_IMAGE"
+	// The environment variable enabling upload of trained model into s3 bucket
+	uploadTrainedModelEnvVar = "UPLOAD_TRAINED_MODEL"
+	// The environment variable specifying s3 bucket folder path used to store model
+	storageBucketModelPath = "AWS_STORAGE_BUCKET_MODEL_PATH"
```
astefanutti, line 48: Can only one be defined, like `AWS_STORAGE_BUCKET_MODEL_PATH` and the model upload logic triggers when it is set / non-empty.

images/util/mc-cli/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM registry.access.redhat.com/ubi9:latest
+WORKDIR /app
+USER 1001
+
+RUN curl https://dl.min.io/client/mc/release/linux-amd64/mc -o /app/mc
```
astefanutti, line 5: Is it the only tool that can be used to push to S3 compatible storage?

sutaakar: well, it is the first tool which I found
for the tests use case I took it as good enough for us

sutaakar: my consideration is that we can reuse the image to push to Minio registry too

tests/kfto/core/environment.go
sutaakar: sure, will update the code

sutaakar: Done, using `AWS_DEFAULT_ENDPOINT` to identify whether to upload model into s3 bucket

images/util/mc-cli/Dockerfile
astefanutti: sounds good.

tests/kfto/core/support.go
```diff
@@ -66,3 +69,95 @@ func PytorchJobCondition(job *kftov1.PyTorchJob, conditionType kftov1.JobConditi
 	}
 	return corev1.ConditionUnknown
 }
+
+func GetOrCreateTestNamespace(t Test) string {
```
abhijeet-dhumal, line 73: Could this `GetOrCreateTestNamespace` function be added/made common for odh tests too?

sutaakar: Does it have sense to have this in codeflare-common?
I put it here as I treat it as a specific use case - for QLoRA tests I need to use existing namespace with PVC containing quantized models, for other approaches the generated namespace is enough. To unify testing approach of "base" KFTO tests with GPU tests I decided to create this helper method to avoid code duplication.

abhijeet-dhumal: Yes, it would be great to have it in codeflare-common as it will be really helpful in local testing too !

sutaakar: Feel free to open a PR moving the function to codeflare-common

PR #246, opened 2024-10-16
Adding ROCm base runtime image for training

Fiona-Waters:
## Description

[RHOAIENG-13241](https://issues.redhat.com/browse/RHOAIENG-13241)
Added a base image for training that includes:
* UBI 9
* Python 3.11
* ROCm 6.1
* PyTorch 2.4.1

## Merge criteria:


images/runtime/training/rocm/Pipfile
```diff
@@ -0,0 +1,34 @@
...
+verify_ssl = true
+name = "pypi"
+
+[packages]
+peft = "*"
+datasets = ">=2.15.0"
+transformers = ">=4.41.2"
+numpy = "<2.0.0,>=1.23.5"
+accelerate = ">=0.34.2"
+torch = "==2.4.1"
```
astefanutti, line 12: The ROCm enabled PyTorch should usual come from a specific index URL, is it the case here?

Fiona-Waters: I wasn't aware of that. Let me look into it. Thanks

Fiona-Waters: torch 2.4.1 is compatible with running `pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm6.` It also installs pillow-10.2.0 torchaudio-2.4.1+rocm6.1 and torchvision-0.19.1+rocm6.1 so maybe we should also include those. wdyt?

Fiona-Waters: Do we want to install it like this, or include it in Pipfile?

Fiona-Waters: I think we need to include it like the example you showed. Will update.

astefanutti: It's likely there is a way to do that in the Pipfile, I think that'd be the preferred way.

Fiona-Waters: Updated to include torch from index in pipfile.

astefanutti: Awesome!

astefanutti (approved):
Thanks!

PR #252, opened 2024-10-18
Initial commit for instructlab standalone script

MichaelClifford:
## Description
This PR adds the initial `standalone.py` script for running the [InstructLab](https://github.com/instructlab) process as a set of Kubernetes Jobs with distributed training (Kubeflow Training Operator).  Currently the script assumes valid SDG data exists in an S3 bucket and that their is an appropriate judge model endpoint available. Currently, the script will download the valid SDG data from S3, run phase 1 training, run phase 2 training, run mt_bench eval, run final eval, and then push the trained model back to an s3 bucket.

This PR also includes the current README.md as well as a helper script for pushing the correct data into S3.

### DIFF
--- a/instructlab/standalone/README.md
+++ b/instructlab/standalone/README.md
@@ -1,0 +1,428 @@
+# Standalone Tool Documentation
+
+## Overview
+
+The `standalone.py` script simulates the [InstructLab](https://instructlab.ai/) workflow within a [Kubernetes](https://kubernetes.io/) environment,
+replicating the functionality of a [KubeFlow Pipeline](https://github.com/kubeflow/pipelines). This allows for distributed training and evaluation
+of models without relying on centralized orchestration tools like KubeFlow.
+
+The `standalone.py` tool provides support for fetching generated SDG (Synthetic Data Generation) data from an AWS S3 compatible object store.
+While AWS S3 is supported, alternative object storage solutions such as Ceph, Nooba, and MinIO are also compatible.
+
+## Requirements
+
+The `standalone.py` script is designed to run within a Kubernetes environment. The following requirements must be met:
+* A Kubernetes cluster with the necessary resources to run the InstructLab workflow.
+  * Nodes with GPUs available for training.
+  * [KubeFlow training operator](https://github.com/kubeflow/training-operator) must be running and `PyTorchJob` CRD must be installed.
+  * A [StorageClass](https://kubernetes.io/docs/concepts/storage/storage-classes/) that supports dynamic provisioning with [ReadWriteMany](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) access mode.
+* A Kubernetes configuration that allows access to the Kubernetes cluster.
+  * Both cluster and in-cluster configurations are supported.
+* SDG data generated and uploaded to an object store.
+
+> [!NOTE]
+> The script can be run outside of the cluster from the command-line, but it requires that the user is currently logged into a Kubernetes cluster.
+
+> [!TIP]
+> Check the `show` command to display an example of a Kubernetes Job that runs the script. Run `./standalone.py show`.
+
+### RBAC Requirements when running in a Kubernetes Job
+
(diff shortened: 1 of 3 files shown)


instructlab/standalone/README.md
```diff
@@ -0,0 +1,428 @@
...
+* A Kubernetes cluster with the necessary resources to run the InstructLab workflow.
+  * Nodes with GPUs available for training.
+  * [KubeFlow training operator](https://github.com/kubeflow/training-operator) must be running and `PyTorchJob` CRD must be installed.
+  * A [StorageClass](https://kubernetes.io/docs/concepts/storage/storage-classes/) that supports dynamic provisioning with [ReadWriteMany](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) access mode.
+* A Kubernetes configuration that allows access to the Kubernetes cluster.
+  * Both cluster and in-cluster configurations are supported.
+* SDG data generated and uploaded to an object store.
+
+> [!NOTE]
+> The script can be run outside of the cluster from the command-line, but it requires that the user is currently logged into a Kubernetes cluster.
```
astefanutti, line 24: Would that be useful to copy the Pipenv files in here as well for local execution?

instructlab/standalone/standalone.py
```diff
@@ -0,0 +1,3199 @@
+#!/usr/bin/env python3
```
astefanutti, line 1: Would it make sense to maintain the generation of this file from the pipeline template or we don't care at that stage?

instructlab/standalone/README.md
```diff
@@ -0,0 +1,428 @@
+# Standalone Tool Documentation
```
astefanutti, line 1: nit: I would title it "InstructLab Standalone Tool".

astefanutti (commented):
LGTM overall, just left a couple of high-level comments.

szaher (commented):
do we need to include requirements.txt with the script dependencies [in this section](https://github.com/opendatahub-io/distributed-workloads/pull/252/files#diff-5289f7a698dd6be9e5dcf954973dff7d18126b913c0f6ef066400f16fc02d22dR24)  just in case the user wants to run it outside k8s? 
Also, do we need to mention that the image used has these dependencies?

astefanutti:
@MichaelClifford could you rebase please?

instructlab/standalone/README.md
```diff
@@ -0,0 +1,428 @@
...
+## Overview
+
+The `standalone.py` script simulates the [InstructLab](https://instructlab.ai/) workflow within a [Kubernetes](https://kubernetes.io/) environment,
+replicating the functionality of a [KubeFlow Pipeline](https://github.com/kubeflow/pipelines). This allows for distributed training and evaluation
+of models without relying on centralized orchestration tools like KubeFlow.
+
+The `standalone.py` tool provides support for fetching generated SDG (Synthetic Data Generation) data from an AWS S3 compatible object store.
+While AWS S3 is supported, alternative object storage solutions such as Ceph, Nooba, and MinIO are also compatible.
+
+## Requirements
```
HumairAK, line 12: > A judge model is assumed to be running external to the script. This is used for model evaluation.

should this be under requirements?

astefanutti (approved):
LGTM

Let's merge it and address the comments in follow up PRs.

instructlab/standalone/README.md
MichaelClifford: Yes, we can do that. One thing worth noting though is that the Pipfile includes the KFP package as its needed for the wider repo, but not for the standalone script.

instructlab/standalone/standalone.py
MichaelClifford: This is a good point. This is likely something we will want to keep track of in the future. But something I hope we can skip for now (at least until this first of kind is finalized).

instructlab/standalone/README.md
astefanutti: Noted, we'll keep that in mind to stick to the minimal set of dependencies.

PR #260, opened 2024-11-05
Dynamically create Judge model Secret resource

sutaakar:
## Description

Create Judge model Secret based on information provided in environment variables.
User doesn't have to manually create Secret with all the needed values.
Environment variables used for Secret creation:
- `JUDGE_API_KEY`: Provide API key for Judge model endpoint
- `JUDGE_NAME`: Name of the Judge model
- `JUDGE_ENDPOINT`: Judge model serving endpoint
- `JUDGE_CA_CERT_FROM_OPENSHIFT`: flag indicating that Judge model endpoint use the same certificate as OpenShift CA where the script is running (so that the certificate should be trusted), set to `true`

## How Has This Been Tested?

Manually by running the test against OpenShift cluster with Judge model deployed in the same cluster.

## Merge criteria:

### DIFF
--- a/tests/odh/ilab_dw_rhoai_test.go
+++ b/tests/odh/ilab_dw_rhoai_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package odh
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"strconv"
@@ -68,15 +69,7 @@ func instructlabDistributedTrainingOnRhoai(t *testing.T, numGpus int) {
 		test.T().Skip("SDG_OBJECT_STORE_DATA_KEY is required to download required data to start training.")
 	}
 
-	// judge model details like endpoint, api-key, model-name, ca certs, ...etc should be provided via k8s secret
-	// we need the secret name so the standalone.py script can fetch the details from that secret.
-	judgeServingModelSecret, judgeServingModelSecretExists := GetJudeServingModelSecret()
 	ilabStorageClassName, ilabStorageClassNameExists := GetStorageClassName()
-
-	if !judgeServingModelSecretExists {
-		test.T().Skip("JUDGE_SERVING_MODEL_SECRET judge model details secret is not provided. ")
-	}
-
 	if !ilabStorageClassNameExists {
 		ilabStorageClassName = ILAB_RHELAI_STORAGE_CLASS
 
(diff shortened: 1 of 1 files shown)


tests/odh/ilab_dw_rhoai_test.go
```diff
@@ -397,6 +392,43 @@ func instructlabDistributedTrainingOnRhoai(t *testing.T, numGpus int) {
...
+
+	judgeServingDetails := map[string]string{
+		judgeServingModelApiKeyEnvVar:   judgeServingModelApiKey,
+		judgeServingModelEndpointEnvVar: judgeServingModelEndpoint,
+		judgeServingModelNameEnvVar:     judgeServingModelName,
+	}
+
+	if judgeServingCaCertFromOpenShiftExists && judgeServingCaCertFromOpenShift == "true" {
+		test.T().Logf("Using OpenShift CA as Judge CA certificate")
+		judgeServingDetails[judgeServingCaCertEnvVar] = "kube-root-ca.crt"
```
abhijeet-dhumal, line 422: Incase of openshift cluster, will these secret name and file name always be the same?
IMHO can we let user specifiy these values using judgeServingCaCertEnvVar, judgeServingCaCertCmKeyEnvVar
Otherwise I think JUDGE_CA_CERT, JUDGE_CA_CERT_CM_KEY env variables are not needed in this flow, WDYT?

sutaakar: > In case of openshift cluster, will these secret name and file name always be the same?

Yes
`kube-root-ca.crt` is created by Kubernetes in all namespaces

> IMHO can we let user specify these values using judgeServingCaCertEnvVar, judgeServingCaCertCmKeyEnvVar

That would require certificate to be available in the namespace.
Which leads to one of these options:
- user have to always pass whole certificate using env variable, test will then create CM and refer it accordingly - that looks like an overkill for current usage, though may reconsider it if Judge is hosted under some insecure endpoint outside of OpenShift
- user have to create ConfigMap by itself and pass its name using env variable - that would affect flexibility of the test as it will require a namespace with predefined ConfigMap
- user use existing ConfigMap as Judge server run in the same OpenShift, in such case env var values are always same, why to always pass it?

Currently implemented approach use just one env variable to indicate that Judge server run on same OpenShift, so proper certificate should be used (which is less than suggested two :) with always the same value).
The only use case which is not covered is Judge server running externally behind insecure endpoint, IMHO we can skip it for now and address it if/once it becomes necessary.

abhijeet-dhumal (commented):
/lgtm

PR #281, opened 2024-11-21
Add "ReadExt" method and update KFTO tests

ChughShilpa:
Add "ReadExt" method and update KFTO tests

## Description
This PR adds support for "ReadExt" method which helped in extracting the python training script in go test binaries. 
Also updated the KFTO tests to use the environment variable from codeflare-common repository

## How Has This Been Tested?
Tested the changes manually by running the KFTO test locally

## Merge criteria:

- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [X] The developer has manually tested the changes and verified that the changes work

### DIFF
--- a/tests/kfto/core/support.go
+++ b/tests/kfto/core/support.go
@@ -19,8 +19,10 @@ package core
 import (
 	"embed"
 	"fmt"
+	"os"
 	"time"
 
+	"github.com/onsi/gomega"
 	. "github.com/onsi/gomega"
 	. "github.com/project-codeflare/codeflare-common/support"
 
@@ -38,6 +40,13 @@ func ReadFile(t Test, fileName string) []byte {
 	t.T().Helper()
 	file, err := files.ReadFile(fileName)
 	t.Expect(err).NotTo(HaveOccurred())
+	return file
+}
+
+func ReadFileExt(t Test, fileName string) []byte {
+	t.T().Helper()
+	file, err := os.ReadFile(fileName)
+	t.Expect(err).NotTo(gomega.HaveOccurred())
 	return file
 }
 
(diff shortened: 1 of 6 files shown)


tests/kfto/core/support.go
```diff
@@ -41,6 +43,13 @@ func ReadFile(t Test, fileName string) []byte {
 	return file
 }
 
+func ReadFileExt(t Test, fileName string) []byte {
```
Fiona-Waters, line 46: Is it necessary to or what is the reason for duplicating this function?

Fiona-Waters (commented):
Just one question otherwise /lgtm
Thanks @ChughShilpa

tests/kfto/core/support.go
sutaakar: The difference is that this function provides file content outside of local folder.
Which makes me think... it shouldn't be needed as `hf_llm_training.py` can be fetched from function above if you adjust `go:embed` to accept also python files.

ChughShilpa: Fiona meant the duplication of the `ReadFileExt` method with the same method present in https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/odh/support.go 

I did the duplication since we already have two separate support.go files in different packages. 
As discussed with @Fiona-Waters, we can have a single support.go file somewhere at more central place so that both can use it.
and this can be addressed in a seaparate PR

ChughShilpa:
```
it shouldn't be needed as hf_llm_training.py can be fetched from function above if you adjust go:embed to accept also python files.
```
Yes you are right, it can be done like this too. 
Thanks for the suggestion @sutaakar

ChughShilpa: Created [PR](https://github.com/opendatahub-io/distributed-workloads/pull/282) to adjust `go:embed`

PR #293, opened 2024-12-11
Upstream Konflux pipelines for training images

astefanutti:
Copy the latest versions of Konflux pipelines from downstream fork for the training images,  so changes can be applied here, as for the Ray images (see #242).


Fiona-Waters (approved):
/lgtm

PR #295, opened 2024-12-17
Add MNIST test to run multi-node distributed training using KFTO

abhijeet-dhumal:
## Description

Added MNIST test to run KFTO Multi-node distributed training using NVIDIA(CUDA) and AMD GPUs(ROCm)
[RHOAIENG-14540](https://issues.redhat.com/browse/RHOAIENG-14540)
[RHOAIENG-14541](https://issues.redhat.com/browse/RHOAIENG-14541)

## How Has This Been Tested?

```
export NOTEBOOK_IMAGE=quay.io/modh/odh-generic-data-science-notebook@sha256:7c1a4ca213b71d342a2d1366171304e469da06d5f15710fab5dd3ce013aa1b73 \
CODEFLARE_TEST_RAY_IMAGE=quay.io/rhoai/ray@sha256:b0e129cd2f4cdea7ad7a7859031357ffd9915410551f94fbcb942af2198cdf78    # CUDA image \
FMS_HF_TUNING_IMAGE=quay.io/modh/fms-hf-tuning@sha256:6f98907f9095db72932caa54094438eae742145f4b66c28d15887d5303ff1186
```
For running test using AMD GPUs, use ROCM image : `quay.io/modh/ray@sha256:db667df1bc437a7b0965e8031e905d3ab04b86390d764d120e05ea5a5c18d1b4`

```
go test -v ./tests/kfto -run TestPyTorchJobMnistCpu
go test -v ./tests/kfto -run TestPyTorchJobMnistWithROCm
go test -v ./tests/kfto -run TestPyTorchJobMnistWithCuda
```

Tested manually : 

- Using CPU : Took 5 mins approx.
- Using NVIDIA GPUs : Took 3 mins approx.
- Using AMD GPUs : Took 3 mins approx.

Note: Usually takes more time to pull training image initially

image

## Merge criteria:

### DIFF
--- a/tests/kfto/kfto_mnist_training_test.go
+++ b/tests/kfto/kfto_mnist_training_test.go
@@ -27,24 +27,41 @@ import (
 	. "github.com/project-codeflare/codeflare-common/support"
 
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestPyTorchJobMnistCpu(t *testing.T) {
-	runKFTOPyTorchMnistJob(t, 0, "", GetCudaTrainingImage(), "resources/requirements.txt")
+	runKFTOPyTorchMnistJob(t, 0, 2, "", GetCudaTrainingImage(), "resources/requirements.txt")
 }
 
 func TestPyTorchJobMnistWithCuda(t *testing.T) {
-	runKFTOPyTorchMnistJob(t, 1, "nvidia.com/gpu", GetCudaTrainingImage(), "resources/requirements.txt")
+	runKFTOPyTorchMnistJob(t, 1, 1, "nvidia.com/gpu", GetCudaTrainingImage(), "resources/requirements.txt")
 }
 
 func TestPyTorchJobMnistWithROCm(t *testing.T) {
-	runKFTOPyTorchMnistJob(t, 1, "amd.com/gpu", GetROCmTrainingImage(), "resources/requirements-rocm.txt")
+	runKFTOPyTorchMnistJob(t, 1, 1, "amd.com/gpu", GetROCmTrainingImage(), "resources/requirements-rocm.txt")
 }
 
-func runKFTOPyTorchMnistJob(t *testing.T, numGpus int, gpuLabel string, image string, requirementsFile string) {
+func runKFTOPyTorchMnistJob(t *testing.T, numGpus int, workerReplicas int, gpuLabel string, image string, requirementsFile string) {
 	test := With(t)
+
+	storageClasses, err := test.Client().Core().StorageV1().StorageClasses().List(test.Ctx(), metav1.ListOptions{})
+	test.Expect(err).NotTo(HaveOccurred(), "Failed to list StorageClasses")
+
+	// Verify at least one StorageClass supports RWX
+	foundRWX := false
+	var storageClassWithRWX storagev1.StorageClass
+	for _, sc := range storageClasses.Items {
+		// Check the allowed access modes in the StorageClass annotations
+		if checkStorageClassSupportsRWX(sc) {
+			foundRWX = true
+			storageClassWithRWX = sc
+			break
+		}
+	}
+	test.Expect(foundRWX).To(BeTrue(), "No StorageClass found with RWX access mode")
 
 	// Create a namespace
 	namespace := test.NewTestNamespace()
@@ -67,12 +84,11 @@ func runKFTOPyTorchMnistJob(t *testing.T, numGpus int, gpuLabel string, image st
 		"requirements.txt": requirementsFileName,
 	})
 
-	// Create PVC for trained model
-	outputPvc := CreatePersistentVolumeClaim(test, namespace.Name, "50Gi", corev1.ReadWriteMany)
+	outputPvc := CreatePersistentVolumeClaimWithStorageClass(test, namespace.Name, storageClassWithRWX, "50Gi", corev1.ReadWriteMany)
 	defer test.Client().Core().CoreV1().PersistentVolumeClaims(namespace.Name).Delete(test.Ctx(), outputPvc.Name, metav1.DeleteOptions{})
 
 	// Create training PyTorch job
-	tuningJob := createKFTOPyTorchMnistJob(test, namespace.Name, *config, gpuLabel, numGpus, outputPvc.Name, image)
+	tuningJob := createKFTOPyTorchMnistJob(test, namespace.Name, *config, gpuLabel, numGpus, workerReplicas, outputPvc.Name, image)
 	defer test.Client().Kubeflow().KubeflowV1().PyTorchJobs(namespace.Name).Delete(test.Ctx(), tuningJob.Name, *metav1.NewDeleteOptions(0))
 
 	// Make sure the PyTorch job is running
@@ -85,7 +101,7 @@ func runKFTOPyTorchMnistJob(t *testing.T, numGpus int, gpuLabel string, image st
 
 }
 
-func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.ConfigMap, gpuLabel string, numGpus int, outputPvcName string, baseImage string) *kftov1.PyTorchJob {
+func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.ConfigMap, gpuLabel string, numGpus int, workerReplicas int, outputPvcName string, baseImage string) *kftov1.PyTorchJob {
 	var useGPU = false
 	var backend string
 
@@ -120,7 +136,7 @@ func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.Config
 										"/bin/bash", "-c",
 										fmt.Sprintf(`mkdir -p /tmp/lib && export PYTHONPATH=$PYTHONPATH:/tmp/lib && \
 										pip install --no-cache-dir -r /mnt/files/requirements.txt --target=/tmp/lib && \
-										python /mnt/files/mnist.py --epochs 1 --save-model --backend %s`, backend),
+										python /mnt/files/mnist.py --epochs 1 --save-model --output-path /mnt/output --backend %s`, backend),
 									},
 									VolumeMounts: []corev1.VolumeMount{
 										{
@@ -130,6 +146,10 @@ func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.Config
 										{
 											Name:      "tmp-volume",
 											MountPath: "/tmp",
+										},
+										{
+											Name:      "output-volume",
+											MountPath: "/mnt/output",
 										},
 									},
 								},
@@ -151,16 +171,43 @@ func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.Config
 										EmptyDir: &corev1.EmptyDirVolumeSource{},
 									},
 								},
+								{
+									Name: "output-volume",
+									VolumeSource: corev1.VolumeSource{
+										PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+											ClaimName: outputPvcName,
+										},
+									},
+								},
 							},
 							RestartPolicy: corev1.RestartPolicyOnFailure,
 						},
 					},
 				},
 				"Worker": {
-					Replicas:      Ptr(int32(1)),
+					Replicas:      Ptr(int32(workerReplicas)),
 					RestartPolicy: kftov1.RestartPolicyOnFailure,
 					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							Labels: map[string]string{
+								"app": "kfto-mnist",
+							},
+						},
 						Spec: corev1.PodSpec{
+							Affinity: &corev1.Affinity{
+								PodAntiAffinity: &corev1.PodAntiAffinity{
+									RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
+										{
+											LabelSelector: &metav1.LabelSelector{
+												MatchLabels: map[string]string{
+													"app": "kfto-mnist",
+												},
+											},
+											TopologyKey: "kubernetes.io/hostname",
+										},
+									},
+								},
+							},
 							Containers: []corev1.Container{
 								{
 									Name:            "pytorch",
@@ -170,7 +217,7 @@ func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.Config
 										"/bin/bash", "-c",
 										fmt.Sprintf(`mkdir -p /tmp/lib && export PYTHONPATH=$PYTHONPATH:/tmp/lib && \
 										pip install --no-cache-dir -r /mnt/files/requirements.txt --target=/tmp/lib && \
-										python /mnt/files/mnist.py --epochs 1 --save-model --backend %s`, backend),
+										python /mnt/files/mnist.py --epochs 1 --save-model --output-path /mnt/output --backend %s`, backend),
 									},
 									VolumeMounts: []corev1.VolumeMount{
 										{
@@ -180,6 +227,10 @@ func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.Config
 										{
 											Name:      "tmp-volume",
 											MountPath: "/tmp",
+										},
+										{
+											Name:      "output-volume",
+											MountPath: "/mnt/output",
 										},
 									},
 								},
@@ -199,6 +250,14 @@ func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.Config
 									Name: "tmp-volume",
 									VolumeSource: corev1.VolumeSource{
 										EmptyDir: &corev1.EmptyDirVolumeSource{},
+									},
+								},
+								{
+									Name: "output-volume",
+									VolumeSource: corev1.VolumeSource{
+										PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+											ClaimName: outputPvcName,
+										},
 									},
 								},
 							},
@@ -247,4 +306,41 @@ func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.Config
 	test.T().Logf("Created PytorchJob %s/%s successfully", tuningJob.Namespace, tuningJob.Name)
 
 	return tuningJob
+}
+
+func checkStorageClassSupportsRWX(sc storagev1.StorageClass) bool {
+	// Provisioners like nfs.csi.k8s.io or kubernetes.io/nfs usually support RWX by default.
+	if sc.Provisioner == "nfs.csi.k8s.io" || sc.Provisioner == "kubernetes.io/nfs" {
+		return true
+	}
+	return false
+}
+
+func CreatePersistentVolumeClaimWithStorageClass(t Test, namespace string, storageClass storagev1.StorageClass, storageSize string, accessMode ...corev1.PersistentVolumeAccessMode) *corev1.PersistentVolumeClaim {
+	t.T().Helper()
+
+	pvc := &corev1.PersistentVolumeClaim{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "PersistentVolumeClaim",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "pvc-",
+			Namespace:    namespace,
+		},
+		Spec: corev1.PersistentVolumeClaimSpec{
+			AccessModes: accessMode,
+			Resources: corev1.VolumeResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceStorage: resource.MustParse(storageSize),
+				},
+			},
+			StorageClassName: &storageClass.Name,
+		},
+	}
+	pvc, err := t.Client().Core().CoreV1().PersistentVolumeClaims(namespace).Create(t.Ctx(), pvc, metav1.CreateOptions{})
+	t.Expect(err).NotTo(HaveOccurred())
+	t.T().Logf("Created PersistentVolumeClaim %s/%s successfully", pvc.Namespace, pvc.Name)
+
+	return pvc
 }
--- a/tests/kfto/kfto_mnist_training_test.go
+++ b/tests/kfto/kfto_mnist_training_test.go
@@ -33,7 +33,6 @@ import (
 func TestPyTorchJobMnistCpu(t *testing.T) {
 	runKFTOPyTorchMnistJob(t, 0, 2, "", GetCudaTrainingImage(), "resources/requirements.txt")
 }
-
 func TestPyTorchJobMnistWithCuda(t *testing.T) {
 	runKFTOPyTorchMnistJob(t, 1, 1, "nvidia.com/gpu", GetCudaTrainingImage(), "resources/requirements.txt")
 }
@@ -133,7 +132,7 @@ func createKFTOPyTorchMnistJob(test Test, namespace string, config corev1.Config
 										"/bin/bash", "-c",
 										fmt.Sprintf(`mkdir -p /tmp/lib && export PYTHONPATH=$PYTHONPATH:/tmp/lib && \
 										pip install --no-cache-dir -r /mnt/files/requirements.txt --target=/tmp/lib && \
-										python /mnt/files/mnist.py --epochs 1 --save-model --output-path /mnt/output --backend %s`, backend),
+										python /mnt/files/mnist.py --epochs 3 --save-model --output-path /mnt/output --backend %s`, backend),
 									},
 									VolumeMounts: []corev1.VolumeMount{
 										{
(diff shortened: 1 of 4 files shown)


tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,267 @@
...
+	var useGPU = false
+	var backend = ""
+
+	if gpuLabel == "nvidia.com/gpu" && numGpus > 0 {
+		useGPU = true
+		backend = "nccl"
+	}
+	if gpuLabel == "amd.com/gpu" && numGpus > 0 {
+		useGPU = true
+		backend = "mpi"
```
astefanutti, line 100: We should look at using `rccl` with ROCm.

tests/kfto/resources/requirements-rocm.txt
```diff
@@ -0,0 +1,5 @@
+--extra-index-url https://download.pytorch.org/whl/rocm6.1
+torch==2.4.0+rocm6.1
```
astefanutti, line 2: The base image should already have the ROCm PyTorch dependency.

tests/kfto/resources/requirements.txt
```diff
@@ -0,0 +1,4 @@
+torch==2.4.0
```
astefanutti, line 1: The base image should already have the CUDA PyTorch dependency.

tests/kfto/resources/requirements.txt
```diff
@@ -0,0 +1,4 @@
...
+torchvision==0.19.0
+tensorboardx==2.6.2
+minio==7.2.12
```
astefanutti, line 4: Is it actually needed?

tests/kfto/resources/requirements-rocm.txt
```diff
@@ -0,0 +1,5 @@
...
+torchvision==0.19.0
+tensorboardx==2.6.2
+minio==7.2.12
```
astefanutti, line 5: Is it actually needed?

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,269 @@
...
+										},
+										{
+											Name:  "MASTER_PORT",
+											Value: "29500",
+										},
+										{
+											Name:  "WORLD_SIZE",
+											Value: "3",
+										},
+									},
```
astefanutti, line 202: This may be removed?

tests/kfto/kfto_mnist_training_test.go
abhijeet-dhumal: But torch.distributed.Backend doesn't have backend type RCCL :eyes: 
https://github.com/pytorch/pytorch/blob/9275091d6e80dabfe6580e0ece425a1e36a59516/torch/_C/_distributed_c10d.pyi#L300
which is being used here : 
https://github.com/opendatahub-io/distributed-workloads/pull/295/files#diff-c2fa1091107fe57d7fe09606f86ed6810e9400b750018172fa4d6302e836dc32R155

astefanutti: Right, I think to use https://github.com/ROCm/rccl as backend with the PyTorch ROCm build, we can still set `nccl` as communication backend. The ROCm build will use RCCL under the hood.

tests/kfto/resources/requirements.txt
abhijeet-dhumal: Actually in this case of creating virtual env first and then executing mnist script, it needs to be installed explicitly 
Still thinking if there is any other way to execute the script without getting error on installing python packages explicitly 
```
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '_message.abi3.so'
Check the permissions.
```

tests/kfto/kfto_mnist_training_test.go
abhijeet-dhumal: I tried to use MPI backend which is considered suitable for AMD env, but it raises issue 
`RuntimeError: Distributed package doesn't have MPI built in. MPI is only included if you build PyTorch from source on a host that has MPI installed.`
[kfto-mnist-pm7z4-worker-0-pytorch (5).log](https://github.com/user-attachments/files/18176912/kfto-mnist-pm7z4-worker-0-pytorch.5.log)

ChughShilpa:
Is there any specific reason not to use the existing [hf_llm_training.py](https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/kfto/resources/hf_llm_training.py) script which is used in [single node KFTO tests](https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/kfto/kfto_training_test.go)

tests/kfto/kfto_mnist_training_test.go
astefanutti: Yes, it's likely the ROCm build of PyTorch does not support MPI. It's better to configure `nccl` as the communication backend so it should use RCCL.

tests/kfto/resources/requirements-rocm.txt
abhijeet-dhumal: https://github.com/opendatahub-io/distributed-workloads/pull/295#discussion_r1889952088

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,269 @@
...
```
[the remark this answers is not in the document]
abhijeet-dhumal: https://github.com/opendatahub-io/distributed-workloads/pull/295#discussion_r1889955146

abhijeet-dhumal:
> Is there any specific reason not to use the existing [hf_llm_training.py](https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/kfto/resources/hf_llm_training.py) script which is used in [single node KFTO tests](https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/kfto/kfto_training_test.go).
> 
> I think test coverage for multi nodes can be done by parameterizing the existing tests to run on multiple nodes.

@ChughShilpa 
Actually there isn't any specific reason, 
when it came to multi-node training first priority came to my mind was to test existing training-operator examples created for the same : https://github.com/opendatahub-io/training-operator/tree/dev/examples
The pytorch based MNIST evxample was already documented a bit, and I wanted to explore more about the backend types used for [torch.distributed](https://github.com/pytorch/pytorch/blob/main/docs/source/distributed.rst#backends)

tests/kfto/kfto_mnist_training_test.go
abhijeet-dhumal: Is it ok to add torchvision package in base image itself?

abhijeet-dhumal: Even if i use NCCL while running it using AMD GPUs, still it seems it is using CUDA by default :eyes: 
```
Using CUDA
Using distributed PyTorch with nccl backend
World Size: 2. Rank: 1
[W1218 08:18:08.612253579 socket.cpp:697] [c10d] The IPv6 network addresses of (kfto-mnist-7sbt8-master-0, 23456) cannot be retrieved (gai error: -2 - Name or service not known).
[W1218 08:18:09.615455688 socket.cpp:697] [c10d] The IPv6 network addresses of (kfto-mnist-7sbt8-master-0, 23456) cannot be retrieved (gai error: -2 - Name or service not known).
```

astefanutti: Yes, I know it's confusing, but it's very likely it's not using CUDA but ROCm :).

Actually, CUDA should not be present in the runtime if as we use the ROCm base image.

It is that way because ROCm has been made to be compatible with the existing ecosystem, so applications originally written for CUDA can be executed seamlessly with ROCm.

abhijeet-dhumal:
- Updated pip package installation logic to avoid `OSError: Permission error while writing/storing package cache on read-only path`
- Updated backend type to 'NCCL' in case of running MNIST training using CUDA/ROCm pytorch images
- Tested and Verified using AMD & NVIDIA GPUs 

[kfto-mnist-vwcfj-master-0-pytorch.log](https://github.com/user-attachments/files/18195711/kfto-mnist-vwcfj-master-0-pytorch.log)
[kfto-mnist-vwcfj-worker-0-pytorch (3).log](https://github.com/user-attachments/files/18195732/kfto-mnist-vwcfj-worker-0-pytorch.3.log)

image (208)

tests/kfto/resources/mnist.py
```diff
@@ -0,0 +1,224 @@
...
+
+import argparse
+import os
+
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+from tensorboardX import SummaryWriter
```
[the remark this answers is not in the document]
abhijeet-dhumal: Yes, I guess it will work.. let me try!

tests/kfto/resources/mnist.py
```diff
@@ -0,0 +1,224 @@
...
+        batch_size=args.test_batch_size,
+        sampler=DistributedSampler(test_ds),
+    )
+
+    for epoch in range(1, args.epochs + 1):
+        train(args, model, device, train_loader, epoch, writer)
+        test(model, device, test_loader, writer, epoch)
+
+    if args.save_model:
+        torch.save(model.state_dict(), "mnist_cnn.pt")
```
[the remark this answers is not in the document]
abhijeet-dhumal: Do you mean to avoid model saving by all worker nodes and just allowing rank 0 node to save the model to avoid redunduncy or ambiguity ?

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,250 @@
...
+func runKFTOPyTorchMnistJob(t *testing.T, numGpus int, gpuLabel string, image string, requirementsFile string) {
+	test := With(t)
+
+	// Create a namespace
+	namespace := test.NewTestNamespace()
+
+	workingDirectory, err := os.Getwd()
+	test.Expect(err).ToNot(HaveOccurred())
+
+	mnist, err := os.ReadFile(workingDirectory + "/resources/mnist.py")
```
sutaakar, line 55: Can you use `ReadFile` function instead?

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,250 @@
...
+								},
+							},
+							RestartPolicy: corev1.RestartPolicyOnFailure,
+						},
+					},
+				},
+				"Worker": {
+					Replicas:      Ptr(int32(1)),
+					RestartPolicy: kftov1.RestartPolicyOnFailure,
+					Template: corev1.PodTemplateSpec{
```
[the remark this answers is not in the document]
abhijeet-dhumal: I have add interpod anti-affinty rule to avoid worker pod scheduling on same node, please review !

tests/kfto/resources/mnist.py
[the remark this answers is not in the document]
abhijeet-dhumal: Accordingly, I have updated script [a bit](https://github.com/opendatahub-io/distributed-workloads/pull/295/files#diff-c2fa1091107fe57d7fe09606f86ed6810e9400b750018172fa4d6302e836dc32R226) to save checkpoint from rank o node and save the model in PV as an outcome, and also added [checks](https://github.com/opendatahub-io/distributed-workloads/pull/295/files#diff-827f326a183edc75d3ff063da5dc955e6e09919cb0b47fad719b9774fe2eb3cfR52) to utilise PVC with RWX accessmode
Please review!

tests/kfto/resources/mnist.py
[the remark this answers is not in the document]
abhijeet-dhumal: Updated tensorboardx requirement to tensorboard :+1:

tests/kfto/kfto_mnist_training_test.go
abhijeet-dhumal: Yeah, Thanks Karel for pointing that out,.. I have used it now accordingly!

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,341 @@
...
+
+func TestPyTorchJobMnistWithCuda(t *testing.T) {
+	runKFTOPyTorchMnistJob(t, 1, 1, "nvidia.com/gpu", GetCudaTrainingImage(), "resources/requirements.txt")
+}
+
+func TestPyTorchJobMnistWithROCm(t *testing.T) {
+	runKFTOPyTorchMnistJob(t, 1, 1, "amd.com/gpu", GetROCmTrainingImage(), "resources/requirements-rocm.txt")
+}
+
+func runKFTOPyTorchMnistJob(t *testing.T, numGpus int, workerReplicas int, gpuLabel string, image string, requirementsFile string) {
```
astefanutti, line 46: We need to make sure multi-nodes is tested so I'd suggest to add an assertion that check `workerReplicas` is strictly greater than 1.

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,341 @@
...
+	var storageClassWithRWX storagev1.StorageClass
+	for _, sc := range storageClasses.Items {
+		// Check the allowed access modes in the StorageClass annotations
+		if checkStorageClassSupportsRWX(sc) {
+			foundRWX = true
+			storageClassWithRWX = sc
+			break
+		}
+	}
+	test.Expect(foundRWX).To(BeTrue(), "No StorageClass found with RWX access mode")
```
astefanutti, line 63: We may one to remove that if the test only do checkpointing on rank-0 node.

tests/kfto/kfto_mnist_training_test.go
abhijeet-dhumal: Actually If we add workerReplicas count greater than 1, then we need that much number of nodes with GPUs,
For example :
If I provide 2 workerNodeReplicas incase of CUDA/ROCm test then I guess it will need 3 separate cluster nodes with GPUs,.. and for adding this test in downstream test-automation, usually we use 2 gpu nodes for testing in QE QG tests, for which this test should be compatible

abhijeet-dhumal: Somehow I want to adjust this test to run on only 2 GPU nodes :thinking: 
image

tests/kfto/resources/mnist.py
```diff
@@ -0,0 +1,233 @@
+from __future__ import print_function
```
astefanutti, line 1: It seems like it can be removed.

tests/kfto/kfto_mnist_training_test.go
astefanutti: Right, I think that's OK to have only two nodes, as rank 0 also participate in the training. We only need to add the pod anti-affinity to the master node spec.

astefanutti:
/lgtm

Awesome work! Could you please share the output logs for an execution of the test on AMD GPUs so I can make sure everything looks OK? 

Let's wait for @sutaakar approval.

abhijeet-dhumal:
Thank you :blush: 
@astefanutti Here is the log file of the master node : 
[kfto-mnist-cdqjx-master-0-pytorch.log](https://github.com/user-attachments/files/18209956/kfto-mnist-cdqjx-master-0-pytorch.log)

abhijeet-dhumal:
Here is the worker node log file : 
[kfto-mnist-mt9dj-worker-0-pytorch.log](https://github.com/user-attachments/files/18209999/kfto-mnist-mt9dj-worker-0-pytorch.log)

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,294 @@
...
+							},
+							RestartPolicy: corev1.RestartPolicyOnFailure,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if useGPU {
```
sutaakar, line 257: Can you also specify resources for non GPU test?

tests/kfto/resources/mnist.py
```diff
@@ -0,0 +1,231 @@
+import argparse
```
sutaakar, line 1: Can you add a licence header here?

astefanutti:
@abhijeet-dhumal Thanks for the logs. They look good to me. The only thing I've noticed is the following error message, it doesn't seem to be blocking, but that'd be better to fix it:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
datasets 3.1.0 requires fsspec[http]<=2024.9.0,>=2023.1.0, but you have fsspec 2024.12.0 which is incompatible.
numba 0.60.0 requires numpy<2.1,>=1.22, but you have numpy 2.2.0 which is incompatible.
```

tests/kfto/kfto_mnist_training_test.go
abhijeet-dhumal: Yesv :eyes: , Now I see it seems resources addition using `useGPU` conditional statement won't be needed as it will be common for each scenario..
Thanks @sutaakar for reviewing this PR !! :raised_hands:

tests/kfto/resources/mnist.py
abhijeet-dhumal: In the original reference, there isn't any license included : https://github.com/opendatahub-io/training-operator/blob/dev/examples/pytorch/mnist/mnist.py
Do you mean to add license same as : https://github.com/opendatahub-io/distributed-workloads/blob/7e021fedf25770f4eca5fdf268badf00717e396d/tests/odh/resources/mnist.py#L1

abhijeet-dhumal:
Latest test results :

-  Added resources list for CPU based usecase
-  Updated requirements.txt to resolve fsspec/numpy package compatibility issue
-  Added license header in MNIST script

image (209)

- CPU usecase      
        [kfto-mnist-9tsrs-master-0-pytorch.log](https://github.com/user-attachments/files/18225897/kfto-mnist-9tsrs-master-0-pytorch.log)
        [kfto-mnist-9tsrs-worker-0-pytorch.log](https://github.com/user-attachments/files/18225902/kfto-mnist-9tsrs-worker-0-pytorch.log)

- AMD GPU usecase
         [kfto-mnist-xmt6x-master-0-pytorch.log](https://github.com/user-attachments/files/18225907/kfto-mnist-xmt6x-master-0-pytorch.log)
         [kfto-mnist-xmt6x-worker-0-pytorch (1).log](https://github.com/user-attachments/files/18225910/kfto-mnist-xmt6x-worker-0-pytorch.1.log)

tests/kfto/kfto_mnist_training_test.go
abhijeet-dhumal: Oh yes, Great catch there.. 👀 , I missed to add resource list for CPU based usecase..
Thanks @sutaakar for reviewing this PR !! 🙌

tests/kfto/resources/mnist.py
sutaakar: I would rather put there license like in https://github.com/opendatahub-io/distributed-workloads/blob/7e021fedf25770f4eca5fdf268badf00717e396d/tests/odh/mnist_ray_test.go

To keep it same as for other files.

abhijeet-dhumal: I have updated license accordingly, please review!

tests/kfto/resources/requirements-rocm.txt
```diff
@@ -0,0 +1,5 @@
+--extra-index-url https://download.pytorch.org/whl/rocm6.1
+torchvision==0.19.0
+tensorboard==2.18.0
+fsspec[http]==2024.6.1
+numpy<2.1,>=1.22
```
sutaakar, line 5: Can you use some specific numpy version?
That way it would be easier to identify what numpy version was actually used in the test without observing logs.

abhijeet-dhumal: Thanks Karel for pointing that out, Yes I have updated numpy version to 2.0.2, to make it consistent!

sutaakar (commented):
/lgtm

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,308 @@
...
+				Operator: corev1.TolerationOpExists,
+			},
+		}
+		tuningJob.Spec.PyTorchReplicaSpecs["Worker"].Template.Spec.Tolerations = []corev1.Toleration{
+			{
+				Key:      gpuLabel,
+				Operator: corev1.TolerationOpExists,
+			},
+		}
+	} else {
```
ChughShilpa, line 287: Resource list for non-gpu case is already defined in tuning job above, and you are also not updating the value for resources as these are the same values 
of L155-L156

abhijeet-dhumal: Ohh right, I guess I did a cherry-pick while working on another PR by keeping this state as a base and while making changes in base commits I reverted those changes due to base commit mismatch conflicts, but forgot to remove else part ..
Thanks @ChughShilpa  for pointing that out! :raised_hands:

abhijeet-dhumal: I have removed else part as it is not needed anymore, please review!

tests/kfto/kfto_mnist_training_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+func TestPyTorchJobMnistWithCuda(t *testing.T) {
+	runKFTOPyTorchMnistJob(t, 1, 1, "nvidia.com/gpu", GetCudaTrainingImage(), "resources/requirements.txt")
+}
+
+func TestPyTorchJobMnistWithROCm(t *testing.T) {
+	runKFTOPyTorchMnistJob(t, 1, 1, "amd.com/gpu", GetROCmTrainingImage(), "resources/requirements-rocm.txt")
+}
+
+func runKFTOPyTorchMnistJob(t *testing.T, numGpus int, workerReplicas int, gpuLabel string, image string, requirementsFile string) {
```
ChughShilpa, line 45: Since we restrict to using workerReplicas=1 then there we can remove this parameter from "runKFTOPyTorchMnistJob" function

abhijeet-dhumal: Yes, Thanks Shilpa for bringing this question...
Initially I was planning to keep 2 worker nodes for CPU usecase as CPUs are usually available in QG tests resource's limit, but to keep the test consistent for all usecases I kept `workerReplicas` same for each case..
As we know master node also participates in the training process, still to make sure multi-nodes is tested, thinking if we can keep `workerReplicas` greater than 1 at least incase of CPUs
cc: @ChughShilpa @sutaakar ??

ChughShilpa: Does that mean you are not sure whether multi-node is tested for Gpu case ?

abhijeet-dhumal: No doubt, multi-node is being tested for sure :heavy_check_mark: , as Master and worker node is participating in the training process,..

abhijeet-dhumal: While running this same test as part of QG execution it at least needs 2 CPUs, but my concern was.. will 3 CPUs always be available?

ChughShilpa:
@abhijeet-dhumal  I see you set the epochs to 1 in test, IMO we should be using higher number of epochs

abhijeet-dhumal:
> @abhijeet-dhumal I see you set the epochs to 1 in test, IMO we should be using higher number of epochs

For the test, 1 epoch should be enough right? 
Because as we increase number of epoch it will just increase number of training iteration by increasing time of execution..
If it is necessary in any case then I'm ok with increasing epoch count :+1:

ChughShilpa:
> > @abhijeet-dhumal I see you set the epochs to 1 in test, IMO we should be using higher number of epochs
> 
> @ChughShilpa For the test, 1 epoch should be enough right? Because as we increase number of epoch it will just increase number of training iteration by increasing time of execution.. and won't effect outcome of the test as base functionality is already being tested If it is necessary in any case then I'm ok with increasing epoch count 👍

I think setting higher epochs provides benefits in identify resource issues or memory leaks that may arise during prolonged training. Also training the model in a more realistic scenario.

1. We are testing with at least 3 epochs for other tests so far (consistency point of view, that can be ignored, just my preference)
2. Number of Iteration depends on the batch size not the epochs. 
3. Yes, higher epochs increase the execution time, since MNIST is a lightweight, so it will not take much higher time.

tests/kfto/kfto_mnist_training_test.go
ChughShilpa: I think we can set worker replicas to 2 in case of CPU, since it makes no big difference as like GPUs :)

abhijeet-dhumal:
I have tested all below cases:-

Training with:
1. workerReplicas-1, epoch-1  --> 2m 28s
2. workerReplicas-2, epoch-1  --> 2m 49s
3. workerReplicas-2, epoch-2  --> 3m 5s

I will update this test accordingly, Thanks !

image

abhijeet-dhumal:
@sutaakar @ChughShilpa 

CPU (Backend type - GLOO) : (WorkerReplicas-2, Epochs-3)
TestPyTorchJobMnistCpu - 3m 2s 

GPU (backend type - NCCL) : (WorkerReplicas-1, Epochs-3)
TestPytorchjobMnistWithROCm - 3m 41s
TestPytorchjobMnistWithCuda - 2m 27s

Note : Initially incase of pulling training image it takes approx 10/12mins, in image below it took 14m 58s to run  TestPytorchjobMnistWithCuda test

image

ChughShilpa (approved):
/lgtm

PR #300, opened 2025-01-06
Add ValidatingAdmissionPolicy tests for RayClusters and AppWrappers

ChristianZaccaria:
## Description

Part of: https://issues.redhat.com/browse/RHOAIENG-14731 

## How Has This Been Tested?

For context, the [VAP-B resources will be applied by the RHOAI Operator](https://issues.redhat.com/browse/RHOAIENG-16133) based on the OCP version it is installed on.

In the meantime, this can be tested by manually applying the resources:
1. Ensure OCP cluster is `>=v4.17`
2. `oc apply` both resources from [this directory](https://github.com/opendatahub-io/kueue/tree/dev/config/rhoai/ocp-17-addons)
4. Ensure Kueue, CodeFlare-Operator(and AppWrapper controller), and KubeRay are installed.
5. Run `go test -timeout 1m ./tests/odh/validating_admission_policy_test.go ./tests/odh/support.go`.

## Additional Thoughts:
- This test must run on an OCP v4.17 or later cluster. Will discuss with @jiripetrlik to find best way to achieve this.
- Taking into account, eventually the rhoai-operator will be the one applying the VAP-B resources. We could apply them ourselves in the meantime for the tests.
- There are 2 `sleeps` in the tests when updating the VAP resource to allow sufficient time for changes to take effect. Thinking of a better way.

## Merge criteria:


tests/odh/validating_admission_policy_test.go
```diff
@@ -0,0 +1,268 @@
...
+	Testing the 1st alternative behavior with the ValidatingAdmissionPolicyBinding enforcement disabled.
+	**************************************************************************/
+	t.Run("Disable the ValidatingAdmissionPolicy enforcement", func(t *testing.T) {
+		vapb, err := test.Client().Core().AdmissionregistrationV1().ValidatingAdmissionPolicyBindings().Get(test.Ctx(), vapb.Name, metav1.GetOptions{})
+		test.Expect(err).ToNot(HaveOccurred())
+
+		vapb.Spec.PolicyName = "none"
+		_, err = test.Client().Core().AdmissionregistrationV1().ValidatingAdmissionPolicyBindings().Update(test.Ctx(), vapb, metav1.UpdateOptions{})
+		test.Expect(err).ToNot(HaveOccurred())
+		time.Sleep(2 * time.Second) // Wait for the ValidatingAdmissionPolicyBinding to be updated
```
KPostOffice, line 171: You should be able to use `test.Eventually` here and get the VAP and check for an expected updated value

ChristianZaccaria: I've given this a try just now, with the debugger I can see it does find the resource to be updated with the updated value, but the tests that follow fail as if the resource had not yet been updated. Perhaps, k8s needs time to propagate these changes somewhere else too?

go.mod
```diff
@@ -1,18 +1,19 @@
 module github.com/opendatahub-io/distributed-workloads
 
-go 1.21
-
-toolchain go1.21.5
+go 1.22.2
```
jiripetrlik, line 3: Is there any particular reason to update Go because of this test?

ChristianZaccaria: It looks like it is not needed. Looks like it was some configuration issue I had locally. Should I revert to `1.21` or keep the upgrade?

tests/odh/validating_admission_policy_test.go
ChristianZaccaria: I've tried implementing:
```
test.Eventually(func() bool {
  vapb, _ = test.Client().Core().AdmissionregistrationV1().ValidatingAdmissionPolicyBindings().Get(test.Ctx(), vapb.Name, metav1.GetOptions{})
  return vapb.Spec.PolicyName == "none"
}, 5*time.Second, 2*time.Second).Should(BeTrue())
```

The issue found is that the 'retry' never gets executed as the updated value is found immediately to be true:
`}, 5*time.Second, 2*time.Second).Should(BeTrue())
`

Using a `sleep` is providing consistent results.

Bobbins228 (commented):
Tried out these tests a couple of times and the testing examples look sound. 
I am not too sure how we can confirm that the VAP updates correctly rather than using sleep but on the testing side /lgtm

tests/odh/validating_admission_policy_test.go
```diff
@@ -0,0 +1,268 @@
...
+		_, err = test.Client().Core().CoreV1().Namespaces().Update(test.Ctx(), ns, metav1.UpdateOptions{})
+		test.Expect(err).ToNot(HaveOccurred())
+
+		// Apply the ValidatingAdmissionPolicyBinding targetting namespaces with the label 'kueue-managed'
+		vapb, err := test.Client().Core().AdmissionregistrationV1().ValidatingAdmissionPolicyBindings().Get(test.Ctx(), vapb.Name, metav1.GetOptions{})
+		test.Expect(err).ToNot(HaveOccurred())
+
+		vapb.Spec.MatchResources.NamespaceSelector.MatchLabels = map[string]string{"kueue-managed": "true"}
+		_, err = test.Client().Core().AdmissionregistrationV1().ValidatingAdmissionPolicyBindings().Update(test.Ctx(), vapb, metav1.UpdateOptions{})
+		test.Expect(err).ToNot(HaveOccurred())
```
ChristianZaccaria, line 223: @Bobbins228 thanks for taking the time to test Mark!

>Tried out these tests a couple of times and the testing examples look sound.
I am not too sure how we can confirm that the VAP updates correctly rather than using sleep but on the testing side /lgtm

These lines ensure that the update happens without errors. Eventhough we ensure the update does happen, the test runs so quickly that subsequent tests fail unless we allow a couple of seconds for it to 'settle'.

tests/odh/validating_admission_policy_test.go
KPostOffice: What exactly needs to propagated?

ChristianZaccaria: I removed the sleeps to avoid flaky tests in the future. I'm using `Eventually` from gomega to create RayClusters that should eventually be evaluating to True, ensuring changes have been propagated to the VAPB resource.

KPostOffice (approved):
/lgtm

PR #312, opened 2025-02-05
Skip GPU usage validation for kfto pytorch multi-node tests

abhijeet-dhumal:
## Description

- While using 2*g4dn.xlarge GPUs, the prometheus metric's GPU usage value shows min 20%
image (190)

- While using A100 GPUS this usage percentage becomes negligible
- Skipping GPU usage validation because the Multi-node tests are expected to run with high end Gpus (in case of MUlti-GPU scenarios) where the training consumes negligible percentage of GPUs

## Merge criteria:

### DIFF
--- a/tests/kfto/kfto_mnist_training_test.go
+++ b/tests/kfto/kfto_mnist_training_test.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"fmt"
 	"testing"
-	"time"
 
 	kftov1 "github.com/kubeflow/training-operator/pkg/apis/kubeflow.org/v1"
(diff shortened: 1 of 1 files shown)


astefanutti:
@sutaakar are you OK removing the current check altogether?

We can try to find a more robust approach separately.

sutaakar:
@astefanutti The Prometheus check is still done in https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/kfto/kfto_training_test.go#L102
MNIST test is so quick that Promethus may not be able scrape GPU workload. `kfto_training_test.go` provides enough workload for the check.

PR #337, opened 2025-02-27
Adjust fms-hf-tuning tests to use granite-3b-code-base-2k model

sutaakar:
## Description

From fms-hf-tuning 2.6.0 the training script isn't able to process existing bloom model.
So this PR upgrades fms tests to use granite-3b-code-base-2k model.

Another upgrade is to use GPU for training, which should speed up test execution time.
For disconnected environment the granite model can be downloaded from S3 bucket.

## Merge criteria:

### DIFF
--- a/tests/fms/kfto_kueue_sft_test.go
+++ b/tests/fms/kfto_kueue_sft_test.go
@@ -17,8 +17,9 @@ limitations under the License.
 package fms
 
 import (
-	"fmt"
+	"bytes"
 	"testing"
+	"time"
 
 	kftov1 "github.com/kubeflow/training-operator/pkg/apis/kubeflow.org/v1"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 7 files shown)


tests/fms/kfto_kueue_sft_test.go
```diff
@@ -113,11 +133,11 @@ func runPytorchjobWithSFTtrainer(t *testing.T, modelConfigFile string, numGpus i
 		Should(WithTransform(PyTorchJobConditionRunning, Equal(corev1.ConditionTrue)))
 
 	// Make sure the PyTorch job succeed
-	test.Eventually(PyTorchJob(test, namespace, tuningJob.Name), TestTimeoutLong).Should(WithTransform(PyTorchJobConditionSucceeded, Equal(corev1.ConditionTrue)))
+	test.Eventually(PyTorchJob(test, namespace, tuningJob.Name), 30*time.Minute).Should(WithTransform(PyTorchJobConditionSucceeded, Equal(corev1.ConditionTrue)))
```
ChughShilpa, line 136: `TestTimeoutGpuProvisioning` can be used here instead of hardcoding the time to 30 mins

Although the naming convention is not suitable for its purpose here :)

ChughShilpa (approved):
Just a nitpick, otherwise /lgtm

tests/fms/kfto_kueue_sft_test.go
sutaakar: Yes, personally I prefer explicitly declared timeouts.
Using timeout from property would indicate that the values are related.

abhijeet-dhumal (approved):

PR #369, opened 2025-04-01
task(RHOAIENG-21368): update CUDA runtime image to Ray 2.44.1

chipspeak:
## Description
As per title, CUDA runtime image has been updated to ~2.44.0. Links to Nvidia gitlab Dockerfiles updated also.

## How Has This Been Tested?
Changes tested via ROSA cluster notebook. Mnist fashion job ran successfully using new image and ray cluster displayed an appropriately adjusted version number.

## Merge criteria:


kryanbeane (changes requested):
@chipspeak Should the `Pipfile.lock` also be updated? I think it's still 2.35.0

chipspeak:
> @chipspeak Should the `Pipfile.lock` also be updated? I think it's still 2.35.0

I'm not sure just because it'll be a fairly extensive change and with the pipfile present, once the install occurs, the file should be updated. I think anyway. Happy to update it if need be though!

kryanbeane (approved):

chipspeak:
> > @chipspeak Should the `Pipfile.lock` also be updated? I think it's still 2.35.0
> 
> I'm not sure just because it'll be a fairly extensive change and with the pipfile present, once the install occurs, the file should be updated. I think anyway. Happy to update it if need be though!

Updated there, thank you kindly for raising it!

images/runtime/ray/cuda/Pipfile
```diff
@@ -5,7 +5,7 @@ name = "pypi"
 
 [packages]
 numpy = {version = "<2.0.0" }
-"ray[all]" = {version = "~=2.35.0" }
+"ray[all]" = {version = "~=2.44.0" }
```
szaher, line 8: should this be 2.44.1 instead of 2.44.0?

images/runtime/ray/cuda/Dockerfile
```diff
@@ -73,32 +73,32 @@ RUN yum install -y \
...
+ENV NV_CUDA_NSIGHT_COMPUTE_DEV_PACKAGE cuda-nsight-compute-12-2-${NV_CUDA_NSIGHT_COMPUTE_VERSION}
 
 RUN yum install -y \
     make \
     findutils \
-    cuda-command-line-tools-12-1-${NV_CUDA_LIB_VERSION} \
-    cuda-libraries-devel-12-1-${NV_CUDA_LIB_VERSION} \
-    cuda-minimal-build-12-1-${NV_CUDA_LIB_VERSION} \
-    cuda-cudart-devel-12-1-${NV_CUDA_CUDART_DEV_VERSION} \
+    cuda-command-line-tools-12-2-${NV_CUDA_LIB_VERSION} \
```
szaher, line 95: not sure if we will need the cli tools for the runtime image. it's probably useful for debugging.

Can we review these changes if we need to do the same thing for our runtime images?

images/runtime/ray/cuda/Dockerfile
```diff
@@ -1,36 +1,36 @@
 ARG PYTHON_VERSION=311
-ARG IMAGE_TAG=1-77.1726664316
+ARG IMAGE_TAG=1-1743079393
```
szaher, line 2: is this the latest image tag?

images/runtime/ray/cuda/Pipfile
chipspeak: I followed the old use of `~=` which results in `2.44.1` being used from this pipfile as it's the latest compatible version. I can hard code it to `2.44.0` if need be. I just stuck with the use of the previous syntax WDYT?

images/runtime/ray/cuda/Dockerfile
chipspeak: last I checked anyway! https://catalog.redhat.com/software/containers/ubi9/python-311/63f764b03f0b02a2e2d63fff

images/runtime/ray/cuda/Dockerfile
chipspeak: Ah sorry, that's just a carryover from the nvidia dockerfile. I'll drop that!

images/runtime/ray/cuda/Pipfile
szaher: would automatically switch to newer releases if the ray release 2.44.2 for example?
Do we need to use exact version ?

chipspeak: Fair question and yes as I understand it. It will stick to versions of 2.44 but wouldn't move to 2.45 for example. Probably worth the larger discussion about floating versions etc!

images/runtime/ray/cuda/Dockerfile
```diff
@@ -1,36 +1,36 @@
...
+ENV NV_CUDA_CUDART_VERSION 12.2.140-1
 
 COPY cuda.repo-x86_64 /etc/yum.repos.d/cuda.repo
 
 RUN NVIDIA_GPGKEY_SUM=d0664fbbdb8c32356d45de36c5984617217b2d0bef41b93ccecd326ba3b80c87 && \
     curl -fsSL https://developer.download.nvidia.com/compute/cuda/repos/rhel9/${NVARCH}/D42D0685.pub | sed '/^Version/d' > /etc/pki/rpm-gpg/RPM-GPG-KEY-NVIDIA && \
     echo "$NVIDIA_GPGKEY_SUM  /etc/pki/rpm-gpg/RPM-GPG-KEY-NVIDIA" | sha256sum -c --strict -
 
-ENV CUDA_VERSION=12.1.1
+ENV CUDA_VERSION=12.2.2
```
jiripetrlik, line 28: Just a small detail. Shouldn't there be also space instead of "=" to be consistent with other changes?

jiripetrlik (approved):
Looks good to me. I have one small suggestion.

PR #371, opened 2025-04-04
Revert "task(RHOAIENG-21368): update CUDA runtime image to Ray 2.44.1"

chipspeak:
Reverts opendatahub-io/distributed-workloads#369


szaher (approved):
/lgtm
There is some issue with the kuberay sync to `red-hat-data-services` so we are reverting this one tbd later

PR #383, opened 2025-04-28
Implement custom test tags for ODH KFTO tests

sutaakar:
## Description

Run KFTO tests only if (test tier equals test tier env variable if defined) AND OpenShift environment match test requirements.
This is the base of work for running ODH e2e tests using image, providing all the configuration using env variables.

## How Has This Been Tested?

Manually by running e2e tests against various OCP environments.

## Merge criteria:

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@
 * `CODEFLARE_TEST_TIMEOUT_LONG` - Timeout duration for long tasks
 * `CODEFLARE_TEST_RAY_IMAGE` (Optional) - Ray image used for raycluster configuration 
 * `MINIO_CLI_IMAGE` (Optional) - Minio CLI image used for uploading/downloading data from/into s3 bucket
+* `TEST_TIER` (Optional) - Specifies test tier to run, skipping tests which don't belong to specified test tier. Supported test tiers: Smoke, Sanity, Tier1, Tier2, Tier3, Pre-Upgrade and Post-Upgrade.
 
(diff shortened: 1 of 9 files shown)


ChughShilpa (approved):
Thanks

sutaakar:
>Just one query, why you keep order of parameters different in func MultiGpu(accelerator Accelerator, numberOfGpus int) than other functions like func MultiNodeGpu(numberOfNodes int, accelerator Accelerator) ?

My intention is to keep this order of parameters: <number of nodes>, <accelerator type>, <number of GPUs in the node>.
Accelerator type serves as a separator which should distinguish which parameter is related to node numbers, and which describe GPU numbers.

sutaakar:
Another approach would be to encode number of nodes and GPUs numbers/type directly into tag name, though that would end with a pile of tags.
WDYT would be a better approach?

ChughShilpa:
> Another approach would be to encode number of nodes and GPUs numbers/type directly into tag name, though that would end with a pile of tags. WDYT would be a better approach?

Yes this approach will increase the no of tags, but it is good to have separate tags as it will be helpful to run manual tests with tags (which are not part of Smoke, Sanity, Tier tags) and can use these tags in separate jenkins pipeline which we planned for RC verification

abhijeet-dhumal (approved):
/lgtm
Thanks a lot!

abhijeet-dhumal (approved):
/lgtm
Thanks a lot karel!

PR #385, opened 2025-04-29
Update runtime images in LLM fine-tuning with KFTO example

astefanutti:
## How Has This Been Tested?

I've successfully run the example with the new images with both supported NVIDIA GPUs and AMD accelerators.

## Merge criteria:


sutaakar:
Can you also update image mentions in readme?
Also, should other examples be upgraded too?

astefanutti:
> Can you also update image mentions in readme?

I've updated the configurations. Please take a look.
Note these are points in time, so it's OK to leave older records too.

> Also, should other examples be upgraded too?

Absolutely. We can do this on an on-going basis.

sutaakar:
/lgtm
Thanks

Fiona-Waters (approved):

PR #400, opened 2025-05-21
Separate the pytorchjob's validating admission policy test for kfto t…

abhijeet-dhumal:
## Description

- Make validating admission policy test separate for pytorchjob & raycluster/appwrapper in separate odh & kfto
- This was these separate tests can be used for downstream's Ray and training components separately

## Merge criteria:

### DIFF
--- a/tests/kfto/validating_admission_policy_test.go
+++ b/tests/kfto/validating_admission_policy_test.go
@@ -1,0 +1,241 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kfto
+
+import (
+	"encoding/json"
+	"testing"
+
+	kftrainingv1 "github.com/kubeflow/training-operator/pkg/apis/kubeflow.org/v1"
+	. "github.com/onsi/gomega"
+
+	vapv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	kueuev1beta1 "sigs.k8s.io/kueue/apis/kueue/v1beta1"
+	testingpytorchjob "sigs.k8s.io/kueue/pkg/util/testingjobs/pytorchjob"
+
+	. "github.com/opendatahub-io/distributed-workloads/tests/common"
+	. "github.com/opendatahub-io/distributed-workloads/tests/common/support"
+)
+
+// Note: This test must run on an OCP v4.17 or later cluster.
+// The Validating Admission Policy feature gate is GA and enabled by default from OCP v4.17 (k8s v1.30)
+
+var (
+	ns            *corev1.Namespace
+	nsNoLabel     *corev1.Namespace
+	rf            *kueuev1beta1.ResourceFlavor
+	cq            *kueuev1beta1.ClusterQueue
+	lq            *kueuev1beta1.LocalQueue
+	pyt           *kftrainingv1.PyTorchJob
+	vapb          *vapv1.ValidatingAdmissionPolicyBinding
+	vapbCopy      *vapv1.ValidatingAdmissionPolicyBinding
+	pytWithLQName = "pyt-with-lq"
+	pytNoLQName   = "pyt-no-lq"
+)
+
+func TestValidatingAdmissionPolicy(t *testing.T) {
+	test := With(t)
+
+	Tags(t, Sanity)
+
+	// Create a namespace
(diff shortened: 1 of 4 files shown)


tests/kfto/validating_admission_policy_test.go
```diff
@@ -0,0 +1,251 @@
...
+func TestValidatingAdmissionPolicy(t *testing.T) {
+	test := With(t)
+
+	// Register RayCluster types with the scheme
+	err := rayv1.AddToScheme(scheme.Scheme)
+	test.Expect(err).ToNot(HaveOccurred())
+
+	// Register AppWrapper types with the scheme
+	err = awv1beta2.AddToScheme(scheme.Scheme)
+	test.Expect(err).ToNot(HaveOccurred())
```
sutaakar, line 71: What is this needed for?

tests/kfto/validating_admission_policy_test.go
```diff
@@ -0,0 +1,251 @@
...
+	pytNoLQName   = "pyt-no-lq"
+)
+
+const (
+	withLQ = true
+	noLQ   = false
+)
+
+func TestValidatingAdmissionPolicy(t *testing.T) {
+	test := With(t)
```
sutaakar, line 63: Can you add tags?

tests/kfto/validating_admission_policy_test.go
```diff
@@ -0,0 +1,251 @@
...
+					"matchLabels": vapbCopy.Spec.MatchResources.NamespaceSelector.MatchLabels,
+				},
+			},
+		},
+	})
+	_, err := test.Client().Core().AdmissionregistrationV1().ValidatingAdmissionPolicyBindings().Patch(test.Ctx(), vapbCopy.Name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
+	test.Expect(err).ToNot(HaveOccurred())
+}
+
+func createPyTorchJob(test Test, namespaceName string, localQueue bool) error {
```
sutaakar, line 239: Nitpick:
it may be better to split the function into two, i.e. :
`createPyTorchJob(test Test, namespaceName string)`
`createPyTorchJobWithLocalQueue(test Test, namespaceName, localQueueName string)`

That will provide more context for the function and also explicitly define what local queue is used.
(my general rule of thumb - using boolean flag as parameter in function is generally not a good approach, that indicates that the function can be split).

tests/kfto/validating_admission_policy_test.go
```diff
@@ -0,0 +1,251 @@
...
+	t.Run("Default ValidatingAdmissionPolicyBinding", func(t *testing.T) {
+		t.Run("PyTorchJob Tests", func(t *testing.T) {
+			t.Run("PyTorchJob should be admitted with the 'kueue.x-k8s.io/queue-name' label set", func(t *testing.T) {
+				err = createPyTorchJob(test, ns.Name, withLQ)
+				test.Expect(err).ToNot(HaveOccurred())
+				defer test.Client().Kubeflow().KubeflowV1().PyTorchJobs(ns.Name).Delete(test.Ctx(), pyt.Name, metav1.DeleteOptions{})
+			})
+			t.Run("PyTorchJob should not be admitted without the 'kueue.x-k8s.io/queue-name' label set", func(t *testing.T) {
+				err = createPyTorchJob(test, ns.Name, noLQ)
+				test.Expect(err).ToNot(BeNil())
```
sutaakar, line 140: Can you also assert content of the error?

tests/kfto/validating_admission_policy_test.go
abhijeet-dhumal: Ohh, right Apologies this is not needed! 
I guess it is meant for registering the scheme which is necessary so that the client can serialize/deserialize the custom resource correctly..
I have added the kftrainingv1 scheme just now.. Please review!

Thanks a lot for reviewing the PR 🙌

tests/kfto/validating_admission_policy_test.go
abhijeet-dhumal: The `Sanity` tag would work right? I have added it just now..

tests/kfto/validating_admission_policy_test.go
```diff
@@ -0,0 +1,247 @@
...
+		if ns.Labels == nil {
+			ns.Labels = map[string]string{}
+		}
+		ns.Labels["kueue.openshift.io/managed"] = "true"
+		_, err = test.Client().Core().CoreV1().Namespaces().Update(test.Ctx(), ns, metav1.UpdateOptions{})
+		test.Eventually(func() bool {
+			ns, _ = test.Client().Core().CoreV1().Namespaces().Get(test.Ctx(), ns.Name, metav1.GetOptions{})
+			return ns.Labels["kueue.openshift.io/managed"] == "true"
+		}).WithTimeout(10 * time.Second).WithPolling(500 * time.Millisecond).Should(BeTrue())
+		test.Expect(err).ToNot(HaveOccurred())
```
sutaakar, line 183: Would it have sense to label the namespace at the same time as you create it?

tests/kfto/validating_admission_policy_test.go
sutaakar: Is the scheme registration necessary?

abhijeet-dhumal: It seems to be a good practice but is not necessary 🤔 . will remove it!
Should I make all these changes in ODH tests too, I couldn't test upadted ODh test reliably because of unsuccessfull appwrapper controller 😞

sutaakar: yes, I think you can adjust ODH too

ChughShilpa:
As mentioned in the test 
```
Note: This test must run on an OCP v4.17 or later cluster.
```
@abhijeet-dhumal 
Make sure to add a logic in downstream tests to avoid test failures with OCP < v4.17

ChughShilpa (approved):
/lgtm

PR #708, opened 2026-03-30
Added OSU benchmark files and results

Sridhar1030:
## Description
Adds a reproducible OSU Micro-Benchmark setup for MPI performance testing on OpenShift using the Kubeflow Training Operator (v1alpha1 TrainJob API).

CPU benchmarks image osu-benchmark:v14 built on quay.io/aipcc/base-images/cpu:3.4.0-ea.2

OpenMPI compiled from source + OSU 7.3
SSH keypair baked into image at /tmp/ssh/ — Kubeflow v1alpha1 does not inject SSH keys automatically
TCP transport forced (-mca pml ob1 -mca btl tcp,self) to work around libfabric/OFI crashing when pods land on different subnets on OpenShift
GPU benchmarks — image osu-benchmark:v11-cuda built on a CUDA 13.0 training base

Two-stage build: Dockerfile.training (CUDA 13.0 + UCX 1.20.0 + OpenMPI 4.1.6 --with-cuda) → Dockerfile.cuda (OSU 7.4 --enable-cuda)
Runs device-to-device benchmarks (-d cuda D D) on Tesla T4 GPUs

## How Has This Been Tested?

Environment: OpenShift cluster, Kubeflow Training Operator v1alpha1, 2 separate physical CPU nodes, 1x g4dn.12xlarge (4x Tesla T4), NVIDIA GPU Operator v25.3.4

CPU — 9 benchmarks across 2 separate nodes:

oc apply -f mpi-runtime.yaml
oc apply -f osu-trainjob.yaml -n mpi-benchmarks
oc logs -n mpi-benchmarks \
  -l batch.kubernetes.io/job-name=osu-benchmark-test-launcher-0 --tail=-1
Results (TCP transport): latency 404 µs (1B), bandwidth 616 MB/s peak, barrier 401 µs — full output in benchmark-results-internode-v14.txt

GPU — 4 device-to-device benchmarks:

oc apply -f mpi-runtime-gpu.yaml
oc apply -f osu-trainjob-gpu.yaml -n mpi-benchmarks
oc logs -n mpi-benchmarks \
  -l batch.kubernetes.io/job-name=osu-benchmark-gpu-launcher-0 --tail=-1
Results: GPU latency 35.83 µs (1B), bandwidth 1,106 MB/s peak — full output in benchmark-results-gpu-v11.txt

Cross-node verified: intra-pod latency 0.30 µs vs inter-node 404 µs (~1347x), confirming real network traffic is measured.

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added OSU Micro-Benchmarks support with CPU and CUDA container images.
  * Added Kubernetes runtimes for MPI CPU and GPU benchmark execution and sample TrainJob manifests.
  * Included example benchmark result files demonstrating inter-node and GPU-Direct runs.

* **Chores**
  * Added GitHub Actions workflow to build/publish benchmark images.
  * Added Makefile targets to build and push CPU and CUDA benchmark images.
  * Added a container startup helper to support running with arbitrary/non-root UIDs.


benchmarks/osu-benchmarks/Dockerfile
```diff
@@ -0,0 +1,135 @@
+# Stage 1: Build OSU Micro-Benchmarks
+FROM quay.io/aipcc/base-images/cpu:3.4.0-ea.2-1772728237 AS builder
```
sutaakar: These images are not available publicly, can you use some of https://github.com/opendatahub-io/base-containers ?

benchmarks/osu-benchmarks/mpi-runtime-gpu.yaml
```diff
@@ -0,0 +1,113 @@
...
+                          cp /tmp/ssh/id_rsa /tmp/.ssh/ &&
+                          cp /tmp/ssh/config /tmp/.ssh/ 2>/dev/null || true &&
+                          chmod 600 /tmp/.ssh/id_rsa &&
+                          W0=osu-benchmark-gpu-worker-0-0.osu-benchmark-gpu.mpi-benchmarks.svc.cluster.local &&
+                          W1=osu-benchmark-gpu-worker-1-0.osu-benchmark-gpu.mpi-benchmarks.svc.cluster.local &&
+                          echo "$W0 slots=1" > /tmp/hostfile &&
+                          echo "$W1 slots=1" >> /tmp/hostfile &&
+                          echo "Waiting for workers...";
+                          sleep 15;
+                          MPI_SSH="ssh -p 2222 -i /tmp/.ssh/id_rsa -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" &&
```
sutaakar: Is this SSH handling needed? Isn't it addressed in https://github.com/opendatahub-io/distributed-workloads/blob/main/images/runtime/training/py312-cuda130-torch29-openmpi41/uid_entrypoint.sh ?

Sridhar1030: I looked at [uid_entrypoint.sh](https://github.com/opendatahub-io/distributed-workloads/blob/main/images/runtime/training/py312-cuda130-torch29-openmpi41/uid_entrypoint.sh) more closely , the training image already provides UID registration (uid_entrypoint.sh, mpirun_wrapper.sh) and SSH client config (StrictHostKeyChecking no, Port 2222). So I'm thinking of dropping the inline nss_wrapper/passwd setup and redundant SSH flags, using uid_entrypoint.sh as the command prefix instead, and only keeping the benchmark-specific bits inline (SSH key copying, hostfile, host key gen, sshd startup).

One gap , the worker needs AcceptEnv CUDA_VISIBLE_DEVICES in its sshd config, which the training image doesn't have. Should I add that upstream to the training image or keep a minimal inline sshd config for it?

sutaakar: you can add it upstream

benchmarks/osu-benchmarks/Dockerfile
Sridhar1030: Switched to `quay.io/opendatahub/odh-midstream-python-base-3-12` from `opendatahub-io/base-containers`. It's UBI 9 with Python 3.12 and OpenShift-compatible (UID 1001, GID 0). Since the ODH Python base doesn't include OpenMPI, the Dockerfile now builds it from source in the builder stage and COPYs it into the runtime  same approach the CUDA image uses. Tested on the cluster and all 9 CPU benchmarks pass.

benchmarks/osu-benchmarks/mpi-runtime-gpu.yaml
Sridhar1030: Done  using `uid_entrypoint.sh` as the command prefix and dropped the inline nss_wrapper/passwd setup. The only thing still inline on the worker is a minimal sshd config with `AcceptEnv CUDA_VISIBLE_DEVICES`  the training image's sshd config doesn't include that yet. Will add it upstream in a follow-up PR; once that's merged the inline config can be removed.

benchmarks/osu-benchmarks/Dockerfile.training
```diff
@@ -0,0 +1,184 @@
+FROM quay.io/opendatahub/odh-midstream-cuda-base-13-0:268a2d4baec5ed3c3ae09a6cce325fb83622d87b
```
sutaakar: What is the purpose of this file? The content seems similar to existing MPI CUDA image.
Would it be possible to reuse existing image instead of declaring the same logic again?

Sridhar1030: You're right — this was a reference copy of the upstream training image Dockerfile and isn't used by anything. `Dockerfile.cuda` already references the published image (`quay.io/opendatahub/odh-training-cuda130-torch29-openmpi41-py312:latest`) directly. Removed it.

Sridhar1030: This was a local copy of the upstream training image Dockerfile (images/runtime/training/py312-cuda130-torch29-openmpi41/) that I used during development to build the base image locally 
The published upstream image didn't have UCX 1.20.0 at the time, which was needed for CUDA 13 support. Now that Dockerfile.cuda pulls the published upstream image directly, this file is no longer needed. 
Removed it.

benchmarks/osu-benchmarks/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+
+ARG SSH_PORT=2222
+
+USER 0
+
+RUN dnf install -y openssh-server openssh-clients \
+    && dnf clean all
+
+LABEL maintainer="srpillai" \
+      summary="OSU MPI Benchmarks on OpenMPI (OpenShift-compatible)" \
```
sutaakar: Nitpick: 

```suggestion
LABEL summary="OSU MPI Benchmarks on OpenMPI (OpenShift-compatible)" \
```

benchmarks/osu-benchmarks/mpi-runtime-cpu.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: trainer.kubeflow.org/v1alpha1
+kind: ClusterTrainingRuntime
+metadata:
+  name: mpi
+spec:
+  mlPolicy:
+    mpi:
+      mpiImplementation: OpenMPI
+      sshAuthMountPath: /tmp/ssh
```
sutaakar, line 9: Have you tried to set `runLauncherAsNode`?
Would it help to avoid the env var setup below?

sutaakar, line 9: Why not to use `sshAuthMountPath: /home/mpiuser/.ssh` ?
AFAIK that way you can reuse ssh config provided by base image, avoiding SSDH config duplication.

Sridhar1030: The launcher is orchestrator-only both MPI ranks run on separate worker pods to ensure clean inter-node latency/bandwidth measurements. 
Setting `runLauncherAsNode: true` would make the launcher a rank, changing the 2-node benchmark topology and would require reworking the hostfile setup, worker replicas, and mpirun args. 
Do we want that for benchmarking?

sutaakar: It is ok as it is, thanks.

Sridhar1030: Tried `sshAuthMountPath: /home/mpiuser/.ssh` , the operator auto-generates and manages the SSH secret for the launcher but doesn't inject the same secret onto worker pods
Defining explicit volumes on workers results in key mismatches since the operator overrides the launcher's volume with its own auto-generated secret
Sticking with /tmp/ssh and manual secret management for now since it works reliably across both roles. Open to exploring other approaches if you think this is worth pursuing further

sutaakar: ok for me to keep it as it is
just was curious about other possibilities

PR #748, opened 2026-04-13
feat: resolve KFTO training images from RHOAI operator RELATED_IMAGE env vars

sutaakar:
## Summary
- KFTO training image getters now dynamically resolve images from the RHOAI operator pod's `RELATED_IMAGE_*` env vars, ensuring tests use the exact digest-pinned images shipped by the deployed RHOAI version
- Three-level resolution priority: local env var (`TEST_TRAINING_*_IMAGE`) → operator pod `RELATED_IMAGE` → hardcoded default
- Bumps fallback default image tags from `rhoai-3.2` to `rhoai-3.4`

## Test plan
- [x] `go build ./tests/kfto/... ./tests/common/support/...` — compiles cleanly
- [x] `go vet ./tests/kfto/... ./tests/common/support/...` — passes
- [ ] Run a KFTO test on cluster and verify operator RELATED_IMAGE is logged and used
- [ ] Verify local env var override (`TEST_TRAINING_*_IMAGE`) still takes precedence

## Summary by CodeRabbit

* **Updates**
  * Training image tags bumped from rhoai-3.2 to rhoai-3.4 across CUDA and ROCm PyTorch variants.

* **Chores**
  * Test infrastructure refactored to use a consistent test context and unified multi-source image resolution.
  * Added operator-based image lookup to resolve related images from the running operator when environment variables are absent.

* **Tests**
  * Updated tests and helpers to use the new context-aware image resolution.


tests/common/support/config.go
```diff
@@ -92,3 +92,27 @@ func GetExpectedRegistry(test Test) string {
...
+					if env.Name == envVarName {
+						return env.Value, true
+					}
+				}
+			}
+		}
+	}
+
+	return "", false
+}
```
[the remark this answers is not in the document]
sutaakar: An empty RELATED_IMAGE value on the operator pod would be a bug in the operator itself — it shouldn't happen in practice. Silently falling back would mask that problem. Keeping the current behavior so it surfaces immediately if it ever occurs.

tests/kfto/kfto_sft_llm_test.go
```diff
@@ -150,7 +151,7 @@ func kftoSftLlm(t *testing.T, image string, gpu Accelerator, modelName string) {
 	}
 
 	for oldValue, newValue := range requiredChangesInNotebook {
-		t.Logf("Replacing '%s' with '%s'", oldValue, newValue)
+		test.T().Logf("Replacing '%s' with '%s'", oldValue, newValue)
 		updatedNotebookContent = strings.Replace(updatedNotebookContent, oldValue, newValue, -1)
 	}
```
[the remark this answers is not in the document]
sutaakar: Good catch. Removed the replacement logging entirely — it adds noise and no real debugging value since the keys are static and visible in the code. Fixed in the latest push.

tests/common/support/environment.go
```diff
@@ -96,28 +96,43 @@ func GetRayTorchROCmImage() string {
...
+// 1. Local env var (TEST_TRAINING_*_IMAGE)
+// 2. RHOAI operator pod RELATED_IMAGE env var
+// 3. Hardcoded default
+func lookupTrainingImage(test Test, envVar, relatedImageEnvVar, defaultImage string) string {
+	if v, ok := os.LookupEnv(envVar); ok {
+		return v
+	}
+	if v, ok := GetRhoaiOperatorRelatedImage(test, relatedImageEnvVar); ok {
+		test.T().Logf("Using operator RELATED_IMAGE %s: %s", relatedImageEnvVar, v)
+		return v
```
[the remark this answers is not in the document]
sutaakar: fine for me to keep it as a tradeoff to keep things simple
in tests it is expected that the value is always provided

tests/common/support/config.go
```diff
@@ -92,3 +92,27 @@ func GetExpectedRegistry(test Test) string {
...
+
+	pods := GetPods(test, "redhat-ods-operator", metav1.ListOptions{
+		FieldSelector: "status.phase=Running",
+	})
+
+	for _, pod := range pods {
+		if strings.HasPrefix(pod.Name, "rhods-operator-") {
+			for _, container := range pod.Spec.Containers {
+				for _, env := range container.Env {
+					if env.Name == envVarName {
```
abhijeet-dhumal, line 109:
```suggestion
					if env.Name == envVarName && env.Value != "" {
```

abhijeet-dhumal: This will make sure the pods created has image associated and not the empty string

tests/common/support/environment.go
```diff
@@ -96,28 +96,43 @@ func GetRayTorchROCmImage() string {
...
-	return lookupEnvOrDefault(TestTrainingRocmPyTorch28Image, TrainingRocmPyTorch28Image)
+func GetTrainingRocmPyTorch28Image(test Test) string {
+	return lookupTrainingImage(test, TestTrainingRocmPyTorch28Image, RelatedImageTrainingRocmPyTorch28, TrainingRocmPyTorch28Image)
+}
+
+// lookupTrainingImage resolves a training image using three-level priority:
+// 1. Local env var (TEST_TRAINING_*_IMAGE)
+// 2. RHOAI operator pod RELATED_IMAGE env var
+// 3. Hardcoded default
+func lookupTrainingImage(test Test, envVar, relatedImageEnvVar, defaultImage string) string {
```
abhijeet-dhumal, line 127: Nit-pick (Optional) : This function calls test.T().Logf, so without Helper(), log/failure will point here instead of the caller.

```suggestion
func lookupTrainingImage(test Test, envVar, relatedImageEnvVar, defaultImage string) string {
  	test.T().Helper()
```

tests/common/support/config.go
sutaakar: Value is intentionally not checked.
If RHOAI contains env var without value I want it to be caught, not silently skipped.

tests/common/support/environment.go
```diff
@@ -96,28 +96,43 @@ func GetRayTorchROCmImage() string {
...
+	if v, ok := os.LookupEnv(envVar); ok {
+		return v
+	}
+	if v, ok := GetRhoaiOperatorRelatedImage(test, relatedImageEnvVar); ok {
+		test.T().Logf("Using operator RELATED_IMAGE %s: %s", relatedImageEnvVar, v)
+		return v
+	}
+	return defaultImage
```
abhijeet-dhumal, line 137:
```suggestion
	test.T().Logf("Operator RELATED_IMAGE %s not found, using default: %s", relatedImageEnvVar, defaultImage)
	return defaultImage
```

ChughShilpa (approved):
/lgtm

tests/common/support/config.go
abhijeet-dhumal: Ah I see, Thanks Karel for confirming 🙌

abhijeet-dhumal:
/approve
Thanks karel 🙌

PR #751, opened 2026-04-13
Add MPI benchmark with SFT+DDP

ChughShilpa:
Add MPI benchmark with SFT+DDP
Resolves [RHOAIENG-51103](https://redhat.atlassian.net/browse/RHOAIENG-51103)

## Description
- Adds a distributed SFT benchmark using **PyTorch DDP with MPI** as the communications backend
- Model: Qwen 2.5 1.5B-Instruct fine-tuned on GSM8K dataset

## How Has This Been Tested?
- Deployed TrainJob on OpenShift cluster (2 nodes x 2 GPUs)
- DDP wrapper applied successfully with MPI backend
- 200 training steps completed without errors
- Gradient synchronization verified via MPI allreduce across all 4 GPUs
- Consistent step times (~3% variance) confirming stable MPI communication

## Benchmark Results (200 steps)
```
| Metric | Value |
| Total time | 25.3 min |
| Avg step time | 7.52s (post-warmup) |
| Min / Max step time | 7.40s / 7.63s |
| Avg throughput | 1.06 samples/s |
| Peak GPU memory | 37.8 GB |
| Avg loss (post-warmup)| 0.4135 |
```

## Merge criteria:
- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has manually tested the changes and verified that the changes work

## Summary by CodeRabbit

* **New Features**
  * Added a runnable MPI+CUDA multi-node supervised fine-tuning benchmark for Qwen 2.5 1.5B using DDP: includes a distributed training script, a cluster runtime manifest, and a job manifest to launch multi-node runs.

* **Documentation**
  * Added a comprehensive README with deployment steps, configurable benchmark and infra parameters (env vars), expected result format, known MPI/CUDA limitations with workarounds, runtime/image details, and cleanup instructions.

### DIFF
--- a/benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
+++ b/benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
@@ -1,0 +1,161 @@
+# MPI DDP SFT Benchmark — Qwen 2.5 1.5B
+
+Distributed Supervised Fine-Tuning benchmark using **PyTorch DDP** with **MPI** as the communications backend, submitted via Kubeflow Trainer v2 `TrainJob`.
+
+## What this benchmark does
+
+| Aspect | Details |
+|--------|---------|
+| Algorithm | SFT with PyTorch DistributedDataParallel (DDP) |
+| Model | Qwen/Qwen2.5-1.5B-Instruct (1.5B params, float32) |
+| Dataset | openai/gsm8k (~7.5 K grade-school math) |
+| Comm backend | **MPI** (`torch.distributed.init_process_group(backend="mpi")`) |
+| Gradient sync | DDP automatic allreduce via MPI |
+| Runtime | `mpi-cuda-openmpi-benchmark` ClusterTrainingRuntime |
+| Image | `quay.io/ksuta/odh-mpi-cuda:0.0.14` |
+
+### MPI communication patterns exercised
+
+- **allreduce** — DDP gradient synchronisation every backward pass
+- **broadcast** — DDP parameter broadcast during initialisation
+- **barrier** — rank-0-first model download coordination
+
+### Why float32
+
+The MPI backend in this image only supports `float32` for CUDA collective operations. `float16` and `bfloat16` fail with `IndexError: map::at` due to missing MPI datatype mappings. See the [Known issues](#known-issues) section for details.
+
+### Why 1.5B model
+
+PyTorch DDP groups all gradients into a single flat buffer for the first training step. For models with more than ~2.1B `float32` parameters, the total element count exceeds MPI's `int32` count limit, causing `MPI_ERR_COUNT: invalid count argument`. The 1.5B model (~1.54B parameters) fits within this limit.
+
+## Files
+
+| File | Description |
+|------|-------------|
+| `train_sft_ddp.py` | Training script — SFT with DDP + MPI gradient allreduce |
+| `trainjob.yaml` | Kubeflow Trainer v2 TrainJob manifest |
+| `mpi-runtime.yaml` | ClusterTrainingRuntime for MPI + CUDA (OpenMPI) |
+
+## Quick Start
+
+### 1. Deploy the ClusterTrainingRuntime
+
+```bash
+oc apply -f mpi-runtime.yaml
+```
+
+### 2. Create namespace and ConfigMap
+
+```bash
+oc create namespace kftv2-mpi-ddp-sft
+
+oc create configmap mpi-ddp-sft-scripts \
+  --from-file=train_sft_ddp.py \
+  -n kftv2-mpi-ddp-sft
+```
+
+### 3. Submit the TrainJob
+
+```bash
+oc create -f trainjob.yaml
+```
+
+### 4. Monitor
+
+```bash
+# Watch pod status
+oc get pods -n kftv2-mpi-ddp-sft -w
+
+# Stream launcher logs (training output)
+oc logs -n kftv2-mpi-ddp-sft \
+  -l batch.kubernetes.io/job-name=<job-name>-launcher-0 \
+  --tail=-1 -f
+```
+
+## Scaling
+
+Adjust `numNodes` and `resourcesPerNode` in `trainjob.yaml` to scale:
+
+| Nodes | GPUs per node | Total GPUs | Use case |
+|------:|:-------------:|-----------:|----------|
+| 2 | 2 | 4 | Default (multi-node verification) |
+| 2 | 4 | 8 | Higher per-node density |
+| 4 | 2 | 8 | More nodes, fewer GPUs each |
+
+## Benchmark parameters
+
+All parameters below are configurable in `trainjob.yaml`. Update them to match your benchmarking requirements.
+
+### Training parameters
+
+Set via `trainer.env` in `trainjob.yaml`. These are injected into all pod containers (launcher and workers) by the Kubeflow Trainer controller.
+
+| Variable | Default | Description | Impact of change |
+|----------|---------|-------------|-----------------|
+| `BENCH_MODEL` | `Qwen/Qwen2.5-1.5B-Instruct` | HuggingFace model ID | Larger model → more GPU memory, slower steps. Must stay under ~2.1B float32 params (MPI int32 limit) |
+| `BENCH_DATASET` | `openai/gsm8k` | HuggingFace dataset ID | Different dataset changes training content but not per-step performance |
+| `BENCH_DATASET_CONFIG` | `main` | Dataset configuration | Selects a subset/split of the dataset |
+| `BENCH_BATCH_SIZE` | `2` | Per-device batch size | Increase → higher GPU utilisation and throughput, more memory. Decrease → less memory, lower throughput |
+| `BENCH_MAX_SEQ_LENGTH` | `512` | Max token sequence length | Increase → more context per example, significantly more memory (attention is O(n²)), slower steps |
+| `BENCH_MAX_STEPS` | `200` | Training steps | Controls total benchmark duration. Does not affect per-step performance |
+| `BENCH_LEARNING_RATE` | `2e-5` | AdamW learning rate | Affects training quality (loss convergence), not throughput |
+| `BENCH_WARMUP_STEPS` | `5` | First N steps excluded from final timing averages | Increase for more accurate steady-state metrics. These steps still execute and log |
+| `BENCH_GRAD_ACCUM` | `1` | Gradient accumulation steps | Increase → simulates larger batch without extra memory, less frequent allreduce |
+| `BENCH_LOG_FREQ` | `1` | Log every N steps | Increase for less verbose output |
+
+### Infrastructure parameters
+
+Set in `trainer` and `podTemplateOverrides` sections of `trainjob.yaml`.
+
+| Parameter | Default | Description | Impact of change |
+|-----------|---------|-------------|-----------------|
+| `numNodes` | `2` | Number of nodes (each runs training + sshd) | More nodes → more GPUs, tests multi-node network scaling |
+| `nvidia.com/gpu` | `"2"` | GPUs per node (also sets MPI processes per node) | More GPUs per node → higher per-node density, more intra-node communication |
+| `memory` | `40Gi` | Memory per node | Increase for larger models or batch sizes |
+| `cpu` | `"8"` | CPU cores per node | Increase if data loading becomes a bottleneck |
+| `dshm sizeLimit` | `16Gi` | Shared memory (`/dev/shm`) for inter-GPU communication | Increase if you see `/dev/shm` out-of-memory errors |
+
+## Expected output
+
+The benchmark prints per-step metrics and a final summary:
+
+```
+============================================================
+BENCHMARK RESULTS (DDP + MPI)
+============================================================
+  Backend:            MPI (via DDP)
+  Model:              Qwen/Qwen2.5-1.5B-Instruct
+  Dtype:              float32
+  World size:         4 GPUs
+  Batch size/GPU:     2
+  Global batch:       8
+  Total steps:        200
+  Avg step time:      X.XXs (post-warmup)
+  Avg throughput:     X.XX samples/s
+  Peak GPU memory:    X.XGB
+============================================================
+```
+
+## Known issues
+
+### MPI backend dtype limitation on CUDA
+
+**Symptom:** `dist.all_reduce` on CUDA tensors fails with `IndexError: map::at` for `bfloat16` / `float16`.
+
+**Root cause:** OpenMPI's MPI datatype table does not include mappings for half-precision formats.
+
+**Workaround:** Load the model with `dtype=torch.float32`.
+
+### MPI int32 count limit (models > ~2B parameters)
+
+**Symptom:** `MPI_ERR_COUNT: invalid count argument` during DDP gradient allreduce on the first training step.
+
+**Root cause:** PyTorch DDP groups all gradients into a single flat buffer. For models with more than ~2.1 billion `float32` parameters, the element count exceeds MPI's `int32` count limit. This is why the benchmark uses the 1.5B model.
+
+## Cleanup
+
+```bash
+oc delete trainjobs -n kftv2-mpi-ddp-sft --all
+oc delete configmap mpi-ddp-sft-scripts -n kftv2-mpi-ddp-sft
+oc delete namespace kftv2-mpi-ddp-sft
+```
--- a/benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/mpi-runtime.yaml
+++ b/benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/mpi-runtime.yaml
@@ -1,0 +1,72 @@
+# ClusterTrainingRuntime for MPI + CUDA (OpenMPI with AIPCC)
+#
+# Deploy:
+#   oc apply -f mpi-runtime.yaml
+---
+apiVersion: trainer.kubeflow.org/v1alpha1
+kind: ClusterTrainingRuntime
+metadata:
+  name: mpi-cuda-openmpi-benchmark
+  labels:
+    trainer.kubeflow.org/framework: openmpi
+spec:
+  mlPolicy:
+    mpi:
+      mpiImplementation: OpenMPI
+      numProcPerNode: 1
+      runLauncherAsNode: true
+      sshAuthMountPath: /home/mpiuser/.ssh
+    numNodes: 1
+  template:
+    spec:
+      network:
+        publishNotReadyAddresses: true
+      replicatedJobs:
+      - groupName: default
+        name: launcher
+        replicas: 1
+        template:
+          metadata:
+            labels:
+              trainer.kubeflow.org/trainjob-ancestor-step: trainer
+          spec:
+            template:
+              spec:
+                containers:
+                - image: quay.io/ksuta/odh-mpi-cuda:0.0.14
+                  name: node
+                  resources:
+                    limits:
+                      nvidia.com/gpu: "1"
+                    requests:
+                      nvidia.com/gpu: "1"
+      - groupName: default
+        name: node
+        replicas: 1
+        template:
+          spec:
+            template:
+              spec:
+                containers:
+                - args:
+                  - -De
+                  - -f
+                  - /home/mpiuser/.sshd_config
+                  command:
+                  - /usr/local/bin/uid_entrypoint.sh
+                  - /usr/sbin/sshd
+                  image: quay.io/ksuta/odh-mpi-cuda:0.0.14
+                  name: node
+                  readinessProbe:
+                    initialDelaySeconds: 3
+                    tcpSocket:
+                      port: 2222
+                  resources:
+                    limits:
+                      nvidia.com/gpu: "1"
+                    requests:
+                      nvidia.com/gpu: "1"
+      successPolicy:
+        operator: All
+        targetReplicatedJobs:
+        - launcher
--- a/benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/train_sft_ddp.py
+++ b/benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/train_sft_ddp.py
@@ -1,0 +1,290 @@
+#!/usr/bin/env python3
+"""
+MPI DDP SFT Benchmark — Qwen 2.5 1.5B + GSM8K
+
+Distributed Supervised Fine-Tuning using PyTorch DistributedDataParallel (DDP)
+with MPI as the communications backend.
+
+Key constraints for MPI + DDP:
+  - Model must use float32 (MPI has no datatype mapping for bfloat16/float16)
+  - Model parameter count must be under ~2.1B (MPI int32 count limit)
+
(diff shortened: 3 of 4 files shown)


benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/train_sft_ddp.py
```diff
@@ -0,0 +1,289 @@
...
+            flush=True,
+        )
+
+    # ---- Tokenizer & Model (rank 0 downloads first to avoid cache races) ----
+    if rank == 0:
+        tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
+        _ = AutoModelForCausalLM.from_pretrained(MODEL_NAME, dtype=torch.float32)
+        del _
+        print("Rank 0: model downloaded and cached", flush=True)
+    dist.barrier()
```
kapil27, line 153: I researched a bit.. `rank == 0` is used for global rank, means for multi-node GPU cluster, it will block the model download on other nodes. 
can you check if `local_rank == 0` can be used instead. that way exactly one process per node will handle the download.
cc. @robert-bell

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+| Aspect | Details |
+|--------|---------|
+| Algorithm | SFT with PyTorch DistributedDataParallel (DDP) |
+| Model | Qwen/Qwen2.5-1.5B-Instruct (1.5B params, float32) |
+| Dataset | openai/gsm8k (~7.5 K grade-school math) |
+| Comm backend | **MPI** (`torch.distributed.init_process_group(backend="mpi")`) |
+| Gradient sync | DDP automatic allreduce via MPI |
+| Runtime | `odh-mpi-cuda-openmpi-aipcc` ClusterTrainingRuntime |
+| Image | `quay.io/ksuta/odh-mpi-cuda:0.0.14` |
```
robert-bell, line 15: @kapil27 do we have the midstream image yet? We should use it.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
```diff
@@ -0,0 +1,229 @@
...
+oc create configmap mpi-ddp-sft-ddp-sshd-config \
+  --from-literal=sshd_config="$(cat <<'EOF'
+PidFile /tmp/sshd.pid
+HostKey /home/mpiuser/.ssh/id_rsa
+StrictModes no
+Port 2222
+AcceptEnv PATH LD_LIBRARY_PATH
+AuthorizedKeysFile /home/mpiuser/.ssh/authorized_keys
+EOF
+)" \
```
robert-bell: is this not automatically configured in the midstream image? @kapil27 ?

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Scaling
+
+Adjust `numNodes` and `resourcesPerNode` in `trainjob.yaml` to scale:
+
+| Nodes | GPUs per node | Total GPUs | Use case |
+|------:|:-------------:|-----------:|----------|
+| 2 | 2 | 4 | Default (multi-node verification) |
+| 2 | 4 | 8 | Higher per-node density |
+| 4 | 2 | 8 | More nodes, fewer GPUs each |
```
robert-bell, line 83: @ChughShilpa have we checked the job can run on 8 GPUs? If not, can we do a quick sanity check? No need to run it to completion. Just make sure it runs a few steps. You'll need to coordinate with the team to make sure no one else is using the gpus.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
```diff
@@ -0,0 +1,229 @@
...
+============================================================
+```
+
+## Known issues and workarounds
+
+### Issue 1: SSH private key permissions on the launcher (client side)
+
+**Symptom:** `mpirun` fails with `Permission denied (publickey)` when connecting to worker nodes.
+
+**Root cause:** The MPI plugin creates a Kubernetes Secret containing the SSH keypair and mounts it as a volume at `/home/mpiuser/.ssh/`. Kubernetes secret volumes default to `0644` file permissions. SSH clients refuse to use a private key (`id_rsa`) that is world-readable — it requires `0600`.
```
robert-bell: @kapil27 is this being fixed upstream? The user and the image shouldn't need to configure this in their scripts. The control plane should handle it.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/train_sft_ddp.py
ChughShilpa: `rank == 0` is intentional here. The barrier ensures at least one process successfully downloads and caches the model before others attempt it. Using local_rank == 0 would allow parallel downloads across nodes, which can hit HuggingFace rate limits. Since each node has its own filesystem, non-rank-0 nodes will still download after the barrier — the purpose is to serialize as a safety check, not to share a cache across nodes.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+```bash
+mpirun --mca plm_rsh_args "-i /tmp/id_rsa -o ConnectionAttempts=10" ...
+```
+
+### Issue 2: sshd authorized_keys path mismatch on worker nodes (server side)
+
+**Symptom:** Node logs show repeated `Connection closed by <IP>` entries. The launcher's SSH readiness poll never succeeds even though sshd is listening on port 2222.
+
+**Root cause:** The container image runs as the `default` user with home directory `/opt/app-root/src` (set by `uid_entrypoint.sh` in `/etc/passwd`). When sshd starts, it resolves `~/.ssh/authorized_keys` to `/opt/app-root/src/.ssh/authorized_keys`. However, the MPI plugin mounts the public key at `/home/mpiuser/.ssh/authorized_keys`. Since these paths don't match, sshd never finds the authorized key and rejects every connection.
```
robert-bell: This shouldn't be necessary no? The configmap should be baked into the image at the right location -- @kapil27 can you check please?

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+```
+AuthorizedKeysFile /home/mpiuser/.ssh/authorized_keys
+```
+
+### Issue 3: Missing /etc/passwd entry on launcher + SSH username mismatch
+
+**Symptom:** SSH authentication succeeds but the connection is immediately reset.
+
+**Root cause:** The `trainer.command` override replaces the container entrypoint, which normally runs `uid_entrypoint.sh` to create an `/etc/passwd` entry. Without it, SSH cannot resolve the current user. Additionally, the node maps the UID to `mpiuser`, but the launcher defaults to `default`.
```
robert-bell: I thought we made a wrapper for mpirun that runs the `uid_entrypoint.sh` script. @kapil27 is that right? Can we fix this by using the right image?

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+**Root cause:** OpenMPI's MPI datatype table does not include mappings for half-precision formats.
+
+**Workaround:** Load the model with `dtype=torch.float32`.
+
+### Issue 6: Environment variable propagation to worker nodes
+
+**Symptom:** Worker nodes load a different model than the launcher (e.g., DDP reports inconsistent parameter counts across ranks).
+
+**Root cause:** `trainer.env` in the TrainJob spec only applies to the launcher pod. Worker nodes launched by `mpirun` do not inherit these environment variables.
```
robert-bell: @kapil27 is this how we have to pass env vars to workers? Does this mean, e.g., the SDK can't pass env vars to the worker pods using `CustomTrainer(env={...})` args? If so we should raise an issue - I'd maybe expect the env to get populated on the workers too.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
```diff
@@ -0,0 +1,229 @@
...
+```bash
+ssh -l mpiuser -i /tmp/id_rsa ...
+mpirun --mca plm_rsh_args "-l mpiuser -i /tmp/id_rsa ..."
+```
+
+### Issue 4: Launcher readiness poll stuck in infinite loop
+
+**Symptom:** `Waiting for worker nodes...` indefinitely.
+
+**Root cause:** The launcher's own hostname appears in `/etc/mpi/hostfile` but it doesn't run sshd.
```
robert-bell: @kapil27 is this a bug that needs fixing in the image? If the runtime or trainjob sets `runLauncherAsNode=true`, then the launcher needs to run `sshd`. Every node needs to run sshd.

How about updating our `mpirun` wrapper so that it also starts sshd?

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/train_sft_ddp.py
```diff
@@ -0,0 +1,289 @@
...
+MODEL_NAME = os.environ.get("BENCH_MODEL", "Qwen/Qwen2.5-1.5B-Instruct")
+DATASET_NAME = os.environ.get("BENCH_DATASET", "openai/gsm8k")
+DATASET_CONFIG = os.environ.get("BENCH_DATASET_CONFIG", "main")
+BATCH_SIZE = int(os.environ.get("BENCH_BATCH_SIZE", "2"))
+MAX_SEQ_LENGTH = int(os.environ.get("BENCH_MAX_SEQ_LENGTH", "512"))
+MAX_STEPS = int(os.environ.get("BENCH_MAX_STEPS", "200"))
+LEARNING_RATE = float(os.environ.get("BENCH_LEARNING_RATE", "2e-5"))
+LOG_FREQ = int(os.environ.get("BENCH_LOG_FREQ", "1"))
+WARMUP_STEPS = int(os.environ.get("BENCH_WARMUP_STEPS", "5"))
+GRADIENT_ACCUMULATION = int(os.environ.get("BENCH_GRAD_ACCUM", "1"))
```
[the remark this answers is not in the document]
ChughShilpa: This is a benchmark script with operator-controlled inputs, not a user-facing API, so the person running it controls the environment variables.

robert-bell (commented):
Thanks @ChughShilpa. I've left a few comments but I think they relate mainly to the image.


Given the time pressure to get this to the PSAP team, I think we shouldn't try fix these problems in this PR. @kapil27 please sanity check this.

The problems do need fixing though before we release - @kapil27 please could you create jira tickets to address the problems? They need fixing before we release.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
kapil27: we are yet to receive midstream image, till then we have to work with this image.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
kapil27: yes, there's an PR: **https://github.com/kubeflow/trainer/issues/3262**

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
ChughShilpa: No, i didn't check yet. But we can check this
Just to metion this script runs on A100 gpus which are avilable in our team cluster

sutaakar:
I tested running an MPI TrainJob against the `odh-mpi-cuda-openmpi-aipcc` ClusterTrainingRuntime with a simple spec (just `command: mpirun` + `args: python /scripts/train.py`) — no SSH workarounds at all — and it completed successfully.

It looks like the workarounds documented in Issues 1–4 of the README and implemented in `trainjob.yaml` are already handled by the ClusterTrainingRuntime:

| Workaround in this PR | Already handled by the runtime |
|---|---|
| `echo "mpiuser:x:$(id -u):0:..." >> /etc/passwd` | `uid_entrypoint.sh` runs as the node container entrypoint |
| `cp id_rsa /tmp && chmod 600` | MPI plugin manages SSH key generation and mounting via `sshAuthMountPath` |
| Custom sshd_config ConfigMap with explicit `AuthorizedKeysFile` | Runtime starts sshd with `-f /home/mpiuser/.sshd_config` |
| SSH readiness polling loop in launcher | Runtime has `readinessProbe` (tcpSocket on port 2222) + Trainer controller handles sequencing |
| `mpirun --mca plm_rsh_args "-l mpiuser -i /tmp/id_rsa"` | MPI plugin injects the correct mpirun arguments |

The `trainjob.yaml` could be significantly simplified by removing these workarounds and relying on the runtime — similar to how a basic TrainJob spec works. The extra `mpi-ddp-sft-ddp-sshd-config` ConfigMap would also no longer be needed.

Were these workarounds developed against an older version of the runtime, or a different runtime?

sutaakar:
**TL;DR:** The SSH workarounds (Issues 1–4) should already be addressed in the `odh-mpi-cuda-openmpi-aipcc` ClusterTrainingRuntime — the TrainJob can be simplified to just `command: mpirun` + `args` without any manual SSH fixes.

sutaakar:
Should the PR also contain ClusterTrainingRuntime?
As right now there is no default provided by Trainer.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
ChughShilpa: It worked with 8 gpus

robert-bell:
Thanks @sutaakar.

> Should the PR also contain ClusterTrainingRuntime?
> As right now there is no default provided by Trainer.

Yes. Let's follow the approach in https://github.com/opendatahub-io/distributed-workloads/pull/708 - @ChughShilpa please can you create the runtime?

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
kapil27: the midstream image does generate a default `/home/mpiuser/.ssdd_config` i.e https://github.com/opendatahub-io/distributed-workloads/blob/6cefb3ef2f6a26ab0b154dbd50e71dc85f8583c0/images/runtime/training/py312-cuda130-torch29-openmpi41/Dockerfile#L135. However it needs investigation on the AutorizedKeysFile.

the midstream image resolved HOME to `opt/app-root/src` for the default user in some OpenShift contexts, sshd looks in the wrong place for the keys. We can eliminate this ConfigMap entirely by opening a quick PR to the runtime image to add `echo "AuthorizedKeysFile /home/mpiuser/.ssh/authorized_keys" >> /home/mpiuser/.sshd_config `in the Dockerfile."

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
kapil27: same as above..  it shouldn't be necesary to mount this via ConfigMap. 

the midstream image does generate a default /home/mpiuser/.ssdd_config i.e https://github.com/opendatahub-io/distributed-workloads/blob/6cefb3ef2f6a26ab0b154dbd50e71dc85f8583c0/images/runtime/training/py312-cuda130-torch29-openmpi41/Dockerfile#L135

However as per the issue raised, it needs investigation on the AutorizedKeysFile.
the midstream image resolved HOME to opt/app-root/src for the default user in some OpenShift contexts, sshd looks in the wrong place for the keys. We can eliminate this ConfigMap entirely by opening a quick PR to the runtime image to add echo "AuthorizedKeysFile /home/mpiuser/.ssh/authorized_keys" >> /home/mpiuser/.sshd_config in the Dockerfile."

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
kapil27: yes, the wrapper handles /etc/password injection i.e: https://github.com/opendatahub-io/distributed-workloads/blob/6cefb3ef2f6a26ab0b154dbd50e71dc85f8583c0/images/runtime/training/py312-cuda130-torch29-openmpi41/mpirun_wrapper.sh#L4

The problem I could think of is https://github.com/opendatahub-io/distributed-workloads/pull/751/changes#diff-19288df484ffcf8644faecb9e1479dc7eb6e5417d673f6938d72f44913d759beR54

is overriding the container ENTRYPOINT, I think we need to investigate whether the bash script is required in the trainjob.yaml or not.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
kapil27: Will research on this.. from my initial investigation.

In the kubeflow/trainer repository, the logic that applies the TrainJob.Spec.Trainer.Env to the underlying pods is located in the JobSet builder plugin: File: [pkg/runtime/framework/plugins/jobset/builder.go]([builder.go](https://github.com/kubeflow/trainer/blob/master/pkg/runtime/framework/plugins/jobset/builder.go)(around line 142) means the operator only injects the Env block into the node (worker) pods if runLauncherAsNode is explicitly set to true in the MPI MLPolicy. If it is false (or not set), Kubernetes will not even inject the environment variables into the worker pod containers.

I have raised the issue upstream for now: here: https://github.com/kubeflow/trainer/issues/3427 will investigate more on this

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
kapil27: the trainjob bash script bypasses the wrapper entirely, also I don't think the launcher needs to run `sshd`. OpenMPI is smart enough to recognize its own hostname/IP in the hostfile. When it needs to spawn processes on the launcher node, it bypasses SSH entirely and uses local fork/exec.

I think the reason this came up as an issue in the benchmark is because our custom bash readiness loop blindly tried to ssh into every single hostname listed in /etc/mpi/hostfile to check if they were awake. It failed when it tried to SSH into itself, because `sshd` wasn't running locally.

c. @sutaakar

kapil27:
I tested the PR as it is and I think it's ready to share with PSAP team. the core architecture works perfectly, I need to investigate thoroughly whether the manual workarounds are actually needed. and as mentioned by @sutaakar , it should work without those workaround. Will update my summary in sometime.

kapil27:
update:  I just tested simplified TrainJob spec on cluster purely relying on `odh-mpi-cuda-openmpi-aipcc` runtime, and it works flawlessly without any of the manual SSH or /etc/passwd workarounds.
@ChughShilpa we need to remove the workarounds and make the trainjob simpler. for the issue you have identified 
regarding "Worker nodes load a different model than the launcher", I have raised an upstream issue: https://github.com/kubeflow/trainer/issues/3427


testing results.
>   Backend:            MPI (via DDP)
>   Model:              Qwen/Qwen2.5-1.5B-Instruct
>   Dtype:              float32
>   World size:         2 GPUs
>   Batch size/GPU:     2
>   Global batch:       4
>   Max seq length:     512
>   Grad accumulation:  1
>   Total steps:        20
>   Total time:         168.2s (2.8 min)
>   Warmup steps:       5
>   Avg step time:      8.39s (post-warmup)
>   Min step time:      8.24s
>   Max step time:      9.19s
>   Avg throughput:     0.48 samples/s
>   Avg loss:           0.4589 (post-warmup)
>   Peak GPU memory:    37.8GB

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/README.md
ChughShilpa: This is yet not fixed in upstream - https://github.com/kubeflow/trainer/pull/3368

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/mpi-runtime.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - groupName: default
+        name: launcher
+        replicas: 1
+        template:
+          spec:
+            template:
+              spec:
+                containers:
+                - env:
+                  - name: OMPI_MCA_mtl
```
kapil27: we can remove these variables, as these are RDMA specific and it will be added by trainjob during runtime if required.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/mpi-runtime.yaml
```diff
@@ -0,0 +1,85 @@
...
+                containers:
+                - args:
+                  - -De
+                  - -f
+                  - /home/mpiuser/.sshd_config
+                  command:
+                  - /usr/local/bin/uid_entrypoint.sh
+                  - /usr/sbin/sshd
+                  env:
+                  - name: OMPI_MCA_mtl
```
kapil27: same here

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/trainjob.yaml
```diff
@@ -0,0 +1,109 @@
...
+    numNodes: 2
+    resourcesPerNode:
+      limits:
+        nvidia.com/gpu: "2"
+        memory: 40Gi
+        cpu: "8"
+      requests:
+        nvidia.com/gpu: "2"
+        memory: 40Gi
+        cpu: "8"
```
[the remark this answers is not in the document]
ChughShilpa: The numProcPerNode: 1 in the runtime is just a default. The controller overrides it based on the TrainJob's resourcesPerNode GPU count.

benchmarks/kftv2-mpi-ddp-sft-ddp/1.5b/train_sft_ddp.py
```diff
@@ -0,0 +1,290 @@
...
+    dist.barrier()
+
+    tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
+    if tokenizer.pad_token is None:
+        tokenizer.pad_token = tokenizer.eos_token
+
+    model = AutoModelForCausalLM.from_pretrained(
+        MODEL_NAME,
+        dtype=torch.float32,
+    ).to(device)
```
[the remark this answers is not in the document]
ChughShilpa: `torch_dtype` is deprecated! Use `dtype` instead! - This was the message showed when used torch_dtype following your suggestion

kapil27 (approved):

kapil27:
/lgtm 
Thanks a lot @ChughShilpa  for this amazing work!

PR #822, opened 2026-05-04
fix: request GPU for failure scenario TrainJob pods

sutaakar:
## Summary
- Add `"nvidia.com/gpu": 1` to `resources_per_node` in the `run_failure_scenario()` helper in `failure_scenarios.ipynb`
- The test is already tagged `Gpu(support.NVIDIA)` but the TrainJob pods were not requesting GPU resources, causing them to land on CPU-only nodes
- On CPU nodes, `instructlab-training` and `unsloth` crash during CUDA init before reaching the intended validation code paths, producing GPU errors instead of the expected `FileNotFoundError` / `No config file found`

## Test plan
- [ ] Run `TestTrainingFailureScenarios` on a cluster with NVIDIA GPUs — all 5 failure scenarios should pass
- [ ] Verify failure scenario pods are scheduled on GPU nodes (`nvidia.com/gpu` in pod resource requests)

## Summary by CodeRabbit

* **Tests**
  * Updated GPU resource configuration in failure scenario testing to ensure comprehensive evaluation across resource environments.

### DIFF
--- a/tests/trainer/resources/failure_scenarios.ipynb
+++ b/tests/trainer/resources/failure_scenarios.ipynb
@@ -145,6 +145,7 @@
     "                resources_per_node={\n",
     "                    \"cpu\": 4,\n",
     "                    \"memory\": \"16Gi\",\n",
+    "                    \"nvidia.com/gpu\": 1,\n",
     "                },\n",
     "            ),\n",
     "            runtime=runtime,\n",


tests/trainer/resources/failure_scenarios.ipynb
```diff
@@ -145,6 +145,7 @@
     "                resources_per_node={\n",
     "                    \"cpu\": 4,\n",
     "                    \"memory\": \"16Gi\",\n",
+    "                    \"nvidia.com/gpu\": 1,\n",
```
MStokluska, line 148: is gpu really needed here? I think I've run these previously and the only problem was image had to be updated to cuda?

sutaakar: "problem" is that unsloth runs some checks from the failure scenarios after verifying that GPU is available.
Without enforcing GPU resource the test Pod can run on CPU, failing on missing GPU instead of expected error.

MStokluska (approved):

PR #837, opened 2026-05-08
feat: add OpenMPI CUDA distributed training tests and notebook

kapil27:
[this will be taken up when MPI is ready for release in RHOAI-3.5]

- Introduced new tests for OpenMPI CUDA distributed training, including integration with Kueue.
- Added a Jupyter notebook (mpi.ipynb) for testing OpenMPI CUDA functionality.
- Enhanced the GenerateNotebookUserToken function to retrieve the token from environment variables if available.

## Merge criteria:

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for distributed GPU training with OpenMPI and CUDA across multiple nodes
  * Added Kueue queue integration tests for distributed training workloads to enable queue management validation
  * Added distributed training validation notebook testing MPI and CUDA functionality in multi-node GPU environments with automatic verification
  * Tests now accept a pre-set notebook token from the environment to skip interactive token retrieval during runs


tests/trainer/sdk_tests/mpi_tests.go
```diff
@@ -0,0 +1,237 @@
...
+							{
+								Name:         corev1.ResourceName(support.NVIDIA.ResourceLabel),
+								NominalQuota: resource.MustParse("2"),
+							},
+						},
+					},
+				},
+			},
+		},
+	})
```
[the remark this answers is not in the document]
kapil27: Thanks. I re-ran this path against the cluster after fixing the queue lifecycle issue, and the notebook workload is admitted on the default queue while the OpenMPI TrainJob is admitted on the explicit LocalQueue as intended. I also added cleanup error logging for the queue resources in this branch. Im keeping the quota sizing aligned with the test workload for now because the end-to-end Kueue MPI SDK test is passing with the current limits.

tests/trainer/sdk_tests/mpi_tests.go
```diff
@@ -0,0 +1,237 @@
...
+		shellQuote(support.GetOpenShiftApiUrl(test)),
+		shellQuote(userToken),
+		shellQuote(namespace.Name),
+		shellQuote(trainerutils.DefaultClusterTrainingRuntimeOpenMPICUDA),
+		queueExport,
+		sdkInstallExports,
+		installKubeflowScript,
+		mpiNotebookName,
+	)
+	command := []string{"/bin/sh", "-c", shellCmd}
```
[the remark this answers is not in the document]
kapil27: Thanks, this is a fair security hardening point in general. I prototyped secret-backed injection for NOTEBOOK_USER_TOKEN, but reverted it for this PR to keep the MPI SDK addition aligned with the existing notebook SDK test pattern already used by the other trainer SDK tests (osft, sft, lora, failure scenarios, etc.), which still export the token into the notebook command path today.

I kept the narrow correctness fix for empty-token handling in GenerateNotebookUserToken(), but I’d prefer to handle the token-delivery hardening as a follow-up that migrates the shared notebook-test path consistently across the whole SDK suite instead of changing only the new MPI path in isolation.

tests/common/environment.go
```diff
@@ -94,6 +94,13 @@ func GetNotebookUserPassword(t Test) string {
 
 // GenerateNotebookUserToken generates an OpenShift token using oc login with username and password
 func GenerateNotebookUserToken(t Test) string {
+	if token, ok := os.LookupEnv(notebookUserToken); ok {
+		if trimmed := strings.TrimSpace(token); trimmed != "" {
+			return trimmed
+		}
+		t.T().Logf("Environment variable %s is set but empty; falling back to oc login", notebookUserToken)
+	}
```
sutaakar, line 104: Nitpick - can you add an info log here saying something like "Environment variable %s is not set, generating Notebook user token from credentials"?
That will help us to see what env variable to set to use it.

tests/trainer/kubeflow_sdk_test.go
```diff
@@ -36,6 +36,18 @@ func TestKubeflowSdkKueueIntegration(t *testing.T) {
...
+func TestKubeflowSdkOpenMPICuda(t *testing.T) {
+	Tags(t, KftoCuda, MultiNodeGpu(2, support.NVIDIA))
+	sdktests.RunOpenMPICudaDistributedTraining(t)
+}
+
+func TestKubeflowSdkOpenMPICudaKueueIntegration(t *testing.T) {
+	Tags(t, KftoCuda, MultiNodeGpu(2, support.NVIDIA))
+	test := support.With(t)
+	support.SetupKueue(test, initialKueueState, support.TrainJobFramework)
+	sdktests.RunOpenMPICudaKueueDistributedTraining(t)
```
sutaakar: Can you pass Accelerator value here, providing it as parameter instead of hardcoding it in implementation functions?

tests/trainer/sdk_tests/mpi_tests.go
```diff
@@ -0,0 +1,243 @@
...
+		mpiNotebookName,
+		0,
+		rwxPvc,
+		common.ContainerSizeMedium,
+		common.GetRecommendedNotebookImageFromImageStream(test, common.NotebookImageStreamTrainingHubCUDA),
+	)
+
+	defer func() {
+		common.DeleteNotebook(test, namespace)
+		test.Eventually(common.Notebooks(test, namespace), support.TestTimeoutGpuProvisioning).Should(HaveLen(0))
```
sutaakar: `TestTimeoutGpuProvisioning` is quite a long timeout, can short timeout be used, to avoid long hanging in case of an issue?

tests/trainer/resources/mpi.ipynb
```diff
@@ -0,0 +1,206 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1de73960",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
```
sutaakar: Could you rather reuse MPI function from https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/trainer/resources/fashion_mnist_mpi.py to have a real life test example?

sutaakar:
Bug: success marker mismatch in `mpi.ipynb` Cell 5.

The training script (`fashion_mnist_mpi.py:111`) logs:
```
MPI TrainJob test PASSED
```

But the notebook assertion checks for:
```python
if "MPI SDK test PASSED" not in log_text:
```

This causes the notebook to fail even when training completes successfully — both ranks log "MPI TrainJob test PASSED" and the job status is `Complete`, but the marker string doesn't match.

**Fix:** In `mpi.ipynb` Cell 5, change:
```python
if "MPI SDK test PASSED" not in log_text:
```
to:
```python
if "MPI TrainJob test PASSED" not in log_text:
```

sutaakar (approved):

PR #852, opened 2026-05-17
fix: wait for progression capture before suspending TrainJob

sutaakar:
## Summary

- Wait for the operator to capture progression data while training pods are still alive before suspending
- Previously the test suspended first, then waited for progression — but pods were already dead and the operator could no longer poll the metrics endpoint
- Reduce metrics poll interval from 8s to 5s (SDK minimum) to maximize capture window for fast training jobs

## Root cause

`TestRhaiFeaturesCuda` has been failing for 8 consecutive nightly builds. The test suspends the TrainJob after 2 epochs (~13s of GPU training), then waits up to 600s for progression > 0. But by that point, pods are terminated and the operator can no longer poll the HTTP metrics endpoint (port 28080). With an 8-second poll interval and only ~13 seconds of training, the operator often missed capturing any progression data before the pods were killed.

## Changes

1. **Wait for progression before suspend** (`rhai_features_tests.go`): Added an `Eventually()` poll that waits for `progressPercentage > 0` while pods are still running, before triggering the suspend
2. **Reduce poll interval** (`rhai_features.ipynb`): Changed `metrics_poll_interval_seconds` from 8 to 5 (SDK minimum) so the operator polls more frequently during short training runs
3. **Simplify Step 5**: Post-suspend step now reads progression state directly instead of polling with a 600s timeout — progression is already guaranteed captured before suspend

## Test plan

- [x] `go vet` and `go build` pass
- [x] Tested on RHOAI 3.4 cluster with A100 GPUs — suspend/resume flow passes, progression captured successfully before suspend (epoch=1.20, progress=24%)
- [ ] Nightly CI validation on RHOAI 3.5 with latest operator (includes opendatahub-io/trainer#153 fix for 100% finalization)

## Release Notes

* **Tests**
  * Updated progression tracking configuration tests to reflect metrics polling interval of 5 seconds
  * Enhanced checkpoint suspend/resume test flow for improved progress tracking verification

### DIFF
--- a/tests/trainer/resources/rhai_features.ipynb
+++ b/tests/trainer/resources/rhai_features.ipynb
@@ -417,7 +417,7 @@
     "trainer_kwargs[\"enable_progression_tracking\"] = enable_progression\n",
     "if enable_progression:\n",
     "    trainer_kwargs[\"metrics_port\"] = 28080\n",
-    "    trainer_kwargs[\"metrics_poll_interval_seconds\"] = 8\n",
+    "    trainer_kwargs[\"metrics_poll_interval_seconds\"] = 5\n",
     "\n",
     "# Add checkpointing if enabled\n",
     "if enable_checkpoint:\n",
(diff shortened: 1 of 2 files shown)


abhijeet-dhumal:
/approve
Thanks karel 🙌

PR #853, opened 2026-05-17
fix: skip downloading unrelated S3 files in Hub test notebooks

sutaakar:
## Summary
- When `AWS_STORAGE_BUCKET_*_DIR` prefix env vars are not set, Hub test notebooks (osft, sft, lora) walk the entire S3 bucket and download every file — including models and datasets belonging to other tests (distilgpt2, granite-3b, MNIST, fashion-MNIST, etc.)
- Only files matching each test's routing rules (model + dataset) are actually used; the rest are written to a default directory and never read
- This change skips files that don't match any known routing rule when no prefix is configured, avoiding ~7GB+ of unnecessary downloads per Hub test run

## Details
The notebooks already classify every S3 file into "dataset", "model", or "default" buckets. Files routed to "default" are not used by the test. When no S3 prefix is set, these unrelated files are now skipped with a log message instead of downloaded.

The fix is identical across all three notebooks — 3 lines added to each:
```python
else:
    if not prefix:
        print(f"[notebook] Skipping unrelated file: {rel}")
        continue
    dst = os.path.join(DATASET_ROOT_NOTEBOOK, rel)
```

When a prefix **is** set, behavior is unchanged — all files under the prefix are downloaded as before.

## Impact
Each Hub test currently downloads the full `rhoai-dw` bucket (~67 files across distilgpt2, granite-3b, qwen2.5, MNIST, fashion-MNIST, table-gpt, etc.). With this fix, each test only downloads its own model + dataset. Combined with 8 Hub tests running per nightly, this eliminates significant redundant I/O.

## Test plan
- [ ] Run OSFT Hub test — verify it downloads only qwen2.5 model + table-gpt dataset, skips distilgpt2/granite/MNIST
- [ ] Run SFT Hub test — verify same skip behavior
- [ ] Run LoRA Hub test — verify same skip behavior
- [ ] Verify tests still pass end-to-end (model + dataset available, training succeeds)
- [ ] Verify that setting `AWS_STORAGE_BUCKET_OSFT_DIR` still downloads all files under that prefix

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated S3 file download routing behavior across training notebooks. Files that don't match established dataset and model routing rules are now skipped (with logging) when no S3 prefix is provided, instead of being routed to default locations. When an S3 prefix is explicitly configured, the original routing behavior remains unchanged.


Fiona-Waters (commented):
Thanks @sutaakar 
/lgtm

MStokluska (approved):

PR #854, opened 2026-05-18
fix: replace TrainJob suspended condition check with pod phase assert…

sutaakar:
…ion in Kueue preemption test

TrainJobConditionSuspended returns Unknown when Kueue admits workloads faster than the operator can reconcile status (suspends → unsuspends before condition is written). Verify pods are Running instead.

## Merge criteria:

## Summary by CodeRabbit

* **Tests**
  * Improved integration test for workload preemption: verification now waits for exactly one training pod to reach Running (selected by jobset label) before proceeding, yielding more reliable validation of preemption and suspension behavior.

### DIFF
--- a/tests/trainer/trainer_kueue_integration_test.go
+++ b/tests/trainer/trainer_kueue_integration_test.go
@@ -24,6 +24,7 @@ import (
 	trainerv1alpha1 "github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1"
 	. "github.com/onsi/gomega"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


ChughShilpa:
Does this change also need to introduce for trainer_kueue_upgrade tests also - https://github.com/sutaakar/distributed-workloads/blob/main/tests/trainer/trainer_kueue_upgrade_training_test.go#L149-L150

sutaakar:
@ChughShilpa it is not needed for `TestRunUpgradeTrainJob`. In this case the issue is caused by race condition between Trainer and Kueue operators when TrainJob is created (Kueue suspends and unsuspends job while Trainer reconciliation is in progress, causing inconsistent TrainJob status).
In `TestRunUpgradeTrainJob` the TrainJob is already suspended, unsuspending by Kueue should properly set TrainJob status.

ChughShilpa (approved):
/lgtm

PR #929, opened 2026-06-22
Add Ray CUDA 12.9 image with Training Hub support

Fiona-Waters:
## Summary

[RHAISTRAT-1693](https://redhat.atlassian.net/browse/RHAISTRAT-1693): Integrate Training Hub with Ray so that Training Hub algorithms (SFT, OSFT, LoRA, GRPO) can be run on Ray clusters in OpenShift AI.
Jira for the image work: https://redhat.atlassian.net/browse/RHOAIENG-61568

Adds a new Ray CUDA 12.9 image (`2.55.1-py312-cu129-th081`) with training-hub and its runtime dependencies (vllm, verl, CUDA extensions) sourced from the RHOAI 3.3 AIPCC index. The existing CUDA 12.8 image (`2.55.1-py312-cu128`) is kept unchanged.

## What's added

**New image: `2.55.1-py312-cu129-th081`**

- CUDA 12.9.2 toolkit (NCCL 2.27.3, cuDNN 9.10.2.21)
- Ray 2.55.1 + training-hub 0.8.1 and transitive dependencies via Pipfile.lock
- AIPCC CUDA wheels: torch 2.9.0, torchvision, torchaudio, triton, vllm 0.13.0, flash-attn 2.8.3, mamba-ssm, causal-conv1d, xformers
- verl 0.8.0 (installed with `--no-deps` to avoid numpy conflict)
- vllm + verl runtime dependencies (anthropic, compressed-tensors, xgrammar, etc.)
- pyzmq from PyPI (AIPCC wheel expects system libzmq which doesn't exist in UBI9)

**New Tekton pipeline: `ray-2.55.1-py312-cu129-th081-push.yaml`**

- Builds and pushes to `quay.io/modh/ray:2.55.1-py312-cu129-th081`
- Triggers on changes to `images/runtime/ray/cuda/2.55.1-py312-cu129-th081/**`

**Unchanged: `2.55.1-py312-cu128`**

- Existing image, Pipfile, Pipfile.lock, and Tekton pipeline are preserved as-is.

## Pipfile changes (cu129-th081)

- Added: `training-hub==0.8.1`, `transformers>=4.57.6,<5.0`, `kernels>=0.9.0,<0.15`, `unsloth>=2026.1.1`, `einops>=0.8`, `bitsandbytes>=0.47.0`, `liger-kernel>=0.5.10`

## How Has This Been Tested?
The image has been tested by running the example notebooks for all current training hub algorithms in this [PR](https://github.com/red-hat-data-services/red-hat-ai-examples/pull/89)  as well as the `1_cluster_job_client.ipynb` notebook - but I did have to increase the resources for the latter. 

## Merge criteria:

[RHAISTRAT-1693]: https://redhat.atlassian.net/browse/RHAISTRAT-1693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chipspeak:
@Fiona-Waters thanks so much for all your work on this Fiona! A thought occurred to us after yesterday's meeting, if we use this tag and we later decide to build our standard Ray image with the same tag we'll overwrite this. Could we add `training-hub` somewhere in the tag so that they're distinct and never run the risk of overwriting each other? That could be the pattern going forward then? cc @kryanbeane @szaher

Fiona-Waters:
> @Fiona-Waters thanks so much for all your work on this Fiona! A thought occurred to us after yesterday's meeting, if we use this tag and we later decide to build our standard Ray image with the same tag we'll overwrite this. Could we add `training-hub` somewhere in the tag so that they're distinct and never run the risk of overwriting each other? That could be the pattern going forward then? cc @kryanbeane @szaher

Good idea. Updated to include `th081` in this commit  : https://github.com/opendatahub-io/distributed-workloads/pull/929/changes/5ab1441101b700c2659ee715c7754d20a299bc33

chipspeak:
LGTM Fiona thanks so much! I might just hold off on the approval until at least 1 other Ray team member has reviewed it.

pawelpaszki:
Was this image manually tested end-to-end before opening the PR? In particular, have GRPO flows been exercised with `verl==0.8.0` given the `--no-deps` install and the runtime ending up on `numpy` 2.x (while verl pins `numpy<2`)? It would be helpful to add testing notes to the PR body — even a short smoke-test summary (imports, one SFT/LoRA job, one GRPO job on Ray) would give reviewers more confidence.

pawelpaszki:
Could you update the PR description and Docker `LABEL`s to reflect the final `th081` naming? The body still references `2.55.1-py312-cu129` / `ray-2.55.1-py312-cu129-push.yaml`, while the image tag, directory, and Tekton pipeline now use `cu129-th081`. Aligning the description and labels (e.g. `name="ray-ubi9-py312-cu129"`) with the published tag would avoid confusion for consumers.

pawelpaszki:
I noticed `Pipfile.lock` resolves `torch==2.10.0` from PyPI, but the Dockerfile then force-reinstalls `torch==2.9.0` from the AIPCC index. Was generating the lockfile against the AIPCC index (with torch 2.9.0 pinned) considered, so the lock and the runtime stack stay aligned and we avoid installing 2.10.0 first only to swap it out? Happy to hear if there’s a reason the current two-step approach is preferred.

pawelpaszki:
The new `ray-2.55.1-py312-cu129-th081-push.yaml` uses an inline `pipelineSpec` like `ray-2.55.1-py312-cu128-push.yaml`, but it omits the post-build tasks that cu128 runs today — e.g. `clair-scan`, `sast-snyk-check`, `ecosystem-cert-preflight-checks`, and `clamav-scan` (added in `d6de7de`).

Was that intentional? If so, do those checks run elsewhere for this image (similar to the ODH `pipelineRef` setup), or is a lighter pipeline acceptable for this variant?

Fiona-Waters:
> Was this image manually tested end-to-end before opening the PR? In particular, have GRPO flows been exercised with `verl==0.8.0` given the `--no-deps` install and the runtime ending up on `numpy` 2.x (while verl pins `numpy<2`)? It would be helpful to add testing notes to the PR body — even a short smoke-test summary (imports, one SFT/LoRA job, one GRPO job on Ray) would give reviewers more confidence.

@pawelpaszki Thank you for the review. Apologies for missing this info. 
The image has been tested by running the example notebooks for all current training hub algorithms in this [PR](https://github.com/red-hat-data-services/red-hat-ai-examples/pull/89)  as well as the `1_cluster_job_client.ipynb` notebook - I had to increase the resources for the latter. 
I have updated the PR description to include this and to reference the new image name with th081 included.

Fiona-Waters:
> I noticed `Pipfile.lock` resolves `torch==2.10.0` from PyPI, but the Dockerfile then force-reinstalls `torch==2.9.0` from the AIPCC index. Was generating the lockfile against the AIPCC index (with torch 2.9.0 pinned) considered, so the lock and the runtime stack stay aligned and we avoid installing 2.10.0 first only to swap it out? Happy to hear if there’s a reason the current two-step approach is preferred.

Good question. The two-step approach is necessary because pipenv (which generates Pipfile.lock) doesn't support custom indexes like the AIPCC index — it resolves everything from PyPI. Since the AIPCC torch 2.9.0 is a custom CUDA-compiled wheel (build 13) that isn't available on PyPI, the lockfile naturally resolves to the latest PyPI torch (2.10.0) as a transitive dependency.

The Dockerfile then force-reinstalls torch==2.9.0 from the AIPCC index to get the correct ABI-matched build. All the other CUDA extensions from AIPCC (vllm, flash-attn, xformers, triton, mamba-ssm, causal-conv1d) are compiled against this specific torch 2.9.0 build, so ABI compatibility requires the AIPCC wheel rather than the PyPI one.

We could pin torch==2.9.0 in the Pipfile to avoid installing 2.10.0 first, but that would pull the PyPI torch 2.9.0 CPU wheel (not the CUDA one) which still gets overwritten. 

> The new `ray-2.55.1-py312-cu129-th081-push.yaml` uses an inline `pipelineSpec` like `ray-2.55.1-py312-cu128-push.yaml`, but it omits the post-build tasks that cu128 runs today — e.g. `clair-scan`, `sast-snyk-check`, `ecosystem-cert-preflight-checks`, and `clamav-scan` (added in `d6de7de`).
> 
> Was that intentional? If so, do those checks run elsewhere for this image (similar to the ODH `pipelineRef` setup), or is a lighter pipeline acceptable for this variant?

This was not intentional. These have been added in the latest commit. Thanks.

pawelpaszki (approved):

PR #930, opened 2026-06-23
feat: add DspClient scaffold and pipeline e2e test structure

Slowlybomb:
## Summary

- Fix DSP env var typo (`dspRoutURL` -> `dspRouteURL`) in `tests/common/environment.go`
- Implement `DspClient` HTTP wrapper for KFP v2beta1 REST API in `tests/common/support/dsp_client.go`
- Add 11 unit tests for `DspClient` using `httptest` mock servers
- Scaffold `tests/pipelines/` package with SFT test skeleton

Ref: [RHOAIENG-43403](https://redhat.atlassian.net/browse/RHOAIENG-43403)

## Note

The SFT test skeleton references `resources/sft_pipeline.yaml` which does not exist yet — compiled pipeline YAMLs from the pipelines repo are needed before the e2e tests can run.

## Test plan

- [x] `go test ./tests/common/support/... -run TestDsp` — 11 unit tests pass
- [x] `go build ./tests/pipelines/...` — builds clean
- [x] `go vet` — no issues
- [x] `golangci-lint` — 0 issues
- [x] `verify-imports` — pass

Made with [Cursor](https://cursor.com)

[RHOAIENG-43403]: https://redhat.atlassian.net/browse/RHOAIENG-43403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/common/support/dsp_client.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return result.State, nil
+}
+
+// GetRunDetails returns the full JSON response for a run, useful for post-run assertions.
+func (c *DspClient) GetRunDetails(runID string) (map[string]interface{}, error) {
+	url := c.baseURL + "/apis/v2beta1/runs/" + runID
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("new request: %w", err)
```
Fiona-Waters, line 193: This returns immediately on any error, including transient network issues. During a 2-hour pipeline run, a single 503 or connection reset would kill the test. Could we consider adding a retry with backoff for transient errors?

tests/pipelines/pipelines_sft_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	. "github.com/opendatahub-io/distributed-workloads/tests/common"
+	"github.com/opendatahub-io/distributed-workloads/tests/common/support"
+)
+
+const (
+	sftPipelineYAMLPath = "resources/sft_pipeline.yaml"
```
Fiona-Waters, line 30: Does this file exist?

tests/pipelines/pipelines_sft_test.go
```diff
@@ -0,0 +1,86 @@
...
+	test.Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("SFT pipeline run %s did not succeed", runID))
+}
+
+func buildSftDataURI(test support.Test) string {
+	endpoint, endpointOK := support.GetStorageBucketDefaultEndpoint()
+	bucket, bucketOK := support.GetStorageBucketName()
+	prefix, prefixOK := support.GetStorageBucketSftDir()
+
+	if endpointOK && bucketOK && prefixOK && endpoint != "" && bucket != "" && prefix != "" {
+		return fmt.Sprintf("s3://%s/%s", bucket, prefix)
```
Fiona-Waters, line 81: Should the endpoint be included here?

tests/common/environment.go
```diff
@@ -206,3 +210,23 @@ func lookupEnvOrDefault(key, value string) string {
...
+
+func GetDspBearerToken(t Test) string {
+	token, ok := os.LookupEnv(dspBearerToken)
+	if !ok {
+		t.T().Fatalf("Required env var %s not set. Set it to an OpenShift service account token with access to the DSP namespace.", dspBearerToken)
+	}
+	return token
+}
+
+func GetDspNamespace() (string, bool) {
```
Fiona-Waters, line 230: Is this function used?

Fiona-Waters (commented):
Thanks @Slowlybomb Just some first review comments. Let's sync and you can show me how the tests work etc.

tests/pipelines/pipelines_sft_test.go
Slowlybomb: No, not yet. This is called out in the PR description under "Note" - the compiled pipeline YAMLs need to come from the pipelines repo before the e2e test can actually run. I can add a t.Skip guard at the top of the test so it skips with a clear message when the file is missing rather than failing. Is it okay?

tests/common/environment.go
Slowlybomb: It's not used yet. I scaffolded it anticipating the SFT test would need to resolve the DSP namespace, but it turned out not to be needed in the current implementation. I'll remove it.

tests/pipelines/pipelines_sft_test.go
Slowlybomb: You're right, the endpoint is fetched but not included in the URI. The question is how the pipeline consumes S3 config: does it expect the endpoint as a separate pipeline parameter (e.g., phase_01_dataset_man_endpoint), or is it configured elsewhere (env var, DSPA config)? Happy to add it as a param if you can confirm the parameter name the pipeline expects.

Fiona-Waters:
@Slowlybomb having looked again at the jira for this work, I think we should close this PR but you should document this work in the spike outcome document. Does that sound okay to you?

Slowlybomb:
> @Slowlybomb having looked again at the jira for this work, I think we should close this PR but you should document this work in the spike outcome document. Does that sound okay to you?

Closed as part of the comment.

PR #1035, opened 2026-09-22
RHOAIENG-82693: add Trainer TLS secure-serving e2e coverage

abhijeet-dhumal:
## Description

Adds E2E coverage for Trainer secure metrics serving and TLS behavior.
- Verifies TLS certificate, secret mount, RBAC, and Prometheus scraping.
- Validates authenticated and unauthenticated metrics access.
- Tests TLS profile changes and controller recovery.
- Ensures TLS profile restoration is retried safely after the test.

## How Has This Been Tested?

image

## Merge criteria:

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end coverage for secure Trainer metrics serving over HTTPS.
  - Validated authenticated metrics access, Prometheus scraping, certificates, permissions, and service configuration.
  - Added optional coverage for TLS profile changes, restoration, and recovery behavior.
  - Added checks for secure serving availability and successful metrics collection.

- **Documentation**
  - Documented commands, timeouts, and environment settings for secure serving, Prometheus scraping, and opt-in TLS profile watcher tests.


tests/trainer/trainer_secure_serving_test.go
```diff
@@ -0,0 +1,754 @@
...
+var (
+	openShiftAPIServerGVR = schema.GroupVersionResource{
+		Group: "config.openshift.io", Version: "v1", Resource: "apiservers",
+	}
+	serviceMonitorGVR = schema.GroupVersionResource{
+		Group: "monitoring.coreos.com", Version: "v1", Resource: "servicemonitors",
+	}
+)
+
+func TestTrainerSecureServing(t *testing.T) {
```
robert-bell: This test is checking a lot of implementation details so is brittle.

Can we simplify it to this
- port forward the **service** -- whatever the equivalent to `kubectl port-forward -n <odh namespace> service/kubeflow-trainer-controller-manager <random available port>:monitoring-port`.  I'm not sure if you can do this with this with the go client. If you can't - look up the pods for the service like you're doing now.
- call the service without a token, ignoring certs => expect 401
- call the service with an invalid token, ignoring certs => expect 403

The rest (prom rbac, certSecret) is implementation detail and doesn't need testing.

tests/trainer/trainer_secure_serving_test.go
```diff
@@ -0,0 +1,754 @@
...
+	test.T().Logf("Waiting for Prometheus to discover Trainer ServiceMonitor target %s/%s", applicationsNamespace, serviceName)
+
+	var target prometheusapiv1.ActiveTarget
+	test.Eventually(func(g Gomega, ctx context.Context) {
+		result, targetErr := prometheus.Targets(ctx)
+		g.Expect(targetErr).NotTo(HaveOccurred())
+		found := false
+		for _, candidate := range result.Active {
+			if string(candidate.Labels["namespace"]) != applicationsNamespace ||
+				string(candidate.Labels["service"]) != serviceName {
```
robert-bell:
```suggestion
				string(candidate.Labels["service"]) != trainerControllerService {
```

tests/trainer/trainer_secure_serving_test.go
```diff
@@ -0,0 +1,754 @@
...
+			}
+			target = candidate
+			found = true
+			break
+		}
+		g.Expect(found).To(BeTrue(), "Trainer ServiceMonitor target was not discovered by Prometheus")
+		g.Expect(target.Health).To(Equal(prometheusapiv1.HealthGood))
+		g.Expect(target.LastError).To(BeEmpty())
+		g.Expect(target.ScrapeURL).To(HavePrefix("https://"))
+		g.Expect(target.ScrapeURL).To(ContainSubstring(":8443"))
```
robert-bell: this is brittle - the port is an implementation detail. If prometheus was able to scrape the metrics successfully then the port should be correct. 

```suggestion
```

tests/trainer/trainer_secure_serving_test.go
```diff
@@ -0,0 +1,754 @@
...
+	}, 5*time.Minute, 10*time.Second).WithContext(test.Ctx()).Should(Succeed())
+	test.T().Logf("Prometheus returned up=1 and certwatcher metrics for Trainer target")
+}
+
+func TestTrainerTLSProfileWatcher(t *testing.T) {
+	Tags(t, Tier2)
+	test := With(t)
+	if !RunTrainerTLSProfileWatcherE2E() {
+		t.Skipf("set %s=true to run the cluster-wide TLS profile mutation test", TrainerTLSProfileWatcherE2E)
+	}
```
robert-bell: rather than having an env flag, can we do discovery - if the cluster doesn't expose the tls profile then mark this test skipped, otherwise it should always run.

tests/trainer/trainer_secure_serving_test.go
```diff
@@ -0,0 +1,754 @@
...
+					return false, nil
+				}
+				if setErr := unstructured.SetNestedField(current.Object, originalProfileMap, "spec", "tlsSecurityProfile"); setErr != nil {
+					return false, setErr
+				}
+				_, updateErr := resource.Update(ctx, current, metav1.UpdateOptions{})
+				if apierrors.IsConflict(updateErr) {
+					lastErr = updateErr
+					return false, nil
+				}
```
robert-bell: we also need to retry on transient errors. I'd keep this simple and just retry on any error.

tests/trainer/trainer_secure_serving_test.go
```diff
@@ -0,0 +1,754 @@
...
+		t.Skip("OpenShift APIServer tlsSecurityProfile has no type")
+	}
+	newProfile := "Intermediate"
+	if originalType == newProfile {
+		newProfile = "Old"
+	}
+	newProfileSpec := tlsSecurityProfile(newProfile)
+	test.T().Logf("TLS profile watcher test will transition APIServer profile %s -> %s", originalType, newProfile)
+
+	defer func() {
```
robert-bell: nit: can we use `test.T().Cleanup` here please

tests/trainer/trainer_secure_serving_test.go
```diff
@@ -0,0 +1,754 @@
...
+					}
+					deployment := getTrainerDeployment(test, applicationsNamespace)
+					return trainerControllerReady(test, applicationsNamespace, deployment, ctx)
+				},
+			)
+			if recoveryErr != nil {
+				test.T().Errorf("failed waiting for Trainer recovery after restoring TLS profile: %v", recoveryErr)
+			} else {
+				test.T().Logf("Trainer recovered after restoring APIServer TLS profile")
+			}
```
robert-bell: any reason not to just keep this simple?

```suggestion
            test.Expect(recoveryErr).NotTo(HaveOccurred(), "failed waiting for Trainer recovery after restoring TLS profile")
```

tests/trainer/trainer_secure_serving_test.go
robert-bell: we also need to retry on transient errors like coderabbit says.

tests/common/support/environment.go
```diff
@@ -73,6 +73,10 @@ const (
 	// filter which ClusterTrainingRuntimes are exercised in e2e tests.
 	TriggerImageName = "TRIGGER_IMAGE_NAME"
 
+	// Enables the Trainer TLS profile watcher test, which changes the
+	// cluster-wide OpenShift APIServer TLS profile and must be explicitly opted in.
+	TrainerTLSProfileWatcherE2E = "TRAINER_TLS_PROFILE_WATCHER_E2E"
+
```
robert-bell: We should remove this - we shouldn't have lots of toggles enabling/disabling tests as it's difficult to maintain.

We can do discovery to skip the test if the profile api isn't available in the cluster (e.g. plain k8s cluster, or old openshift version).

robert-bell (commented):
thanks @abhijeet-dhumal. Can you split the tests into 2 separate test files - one for the metrics/prometheus bits, another for the tls profile bits please?

PR #1037, opened 2026-09-23
fix: detect ODH builds by subscription channel

ChughShilpa:
Detect the build from the operator subscription channel: `odh-stable` for ODH nightly, `beta` for RHOAI EA, and `stable-x.y` for RHOAI GA. Return an error for unknown channels or missing/ambiguous subscriptions.

Log the build type, release type, channel, and installed CSV. Report both skipped speculator CTR names from the expected runtime list, including when either is absent from the cluster. 

Validation:

- `make unit-test`
- `make golangci-lint 'LINT_PKG=./tests/common/support/... ./tests/trainer/...'`
- `make verify-imports`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Build identification now distinguishes ODH and RHOAI using operator subscription channels. Missing, ambiguous, or unsupported subscriptions now produce an error instead of being treated as a non-RHOAI build.
  * ODH test output now consistently identifies skipped speculator runtimes, and expected CUDA runtime images have been updated.


tests/common/support/rhoai_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+			buildType, err := GetBuildType(test)
+			test.Expect(err).To(gomega.MatchError(gomega.ContainSubstring(tc.wantError)))
+			test.Expect(buildType).To(gomega.BeEmpty())
+		})
+	}
+}
+
+func TestGetBuildTypeUnsupportedChannels(t *testing.T) {
+	for _, channel := range []string{"", "stable", "fast", "stable-next", "stable-3", "stable-3.4.1", "stable-3.4-ea"} {
```
sutaakar: IMHO we shouldn't be detecting unsupported channels, as RHOAI provides many channels and explicitly checking them is fragile. It can be enough to detect ODH channel, if it doesn't match then assume RHOAI.

sutaakar (approved):

